PR #16, opened 2017-11-10
Newsli - The healthier way to digest the news

haletothewood:


js/controllers/newsliController.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+(function (exports) {
+  exports.NewsliController = function (ArticleList, ArticleListView, ArticleView) {
+    articleList = new ArticleList;
+    articleListView = new ArticleListView(articleList)
+    articleView = new ArticleView;
```
mdsomerfield: I would generally recommend instantiation these and then passing in the instances, rather than passing in the classes and then instantiating. The latter pattern is generally only used when you need to generate instances at run time e.g. you have received some information from an API and you want to create a model with that information.

js/controllers/newsliController.js
```diff
@@ -0,0 +1,96 @@
...
+
+    articlesContainer = document.getElementById("articles-view");
+    articleContainer = document.getElementById("single-article-view");
+    imageNodes = document.getElementsByTagName("IMG")
+    backButton = "<button id='back' type='button'>Back<button>"
+
+    makersDomain = 'http://news-summary-api.herokuapp.com/';
+    guardianAPIRequest = 'guardian?apiRequestUrl='
```
mdsomerfield: This kind of info usually goes in a settings or configuration class/file.

js/controllers/newsliController.js
```diff
@@ -0,0 +1,96 @@
...
+          updateArticleList(articles)
+          showSingleArticle()
+        };
+      };
+      xhttp.open("GET", makersDomain + guardianAPIRequest + url + allFields, true);
+      xhttp.send()
+    }
+
+    function showSingleArticle() {
+      imageArray = [].slice.call(imageNodes)
```
mdsomerfield: Took me a while to follow the flow here. The code itself is clean and readable which is great, but I think that you're trying to do too many things in the same class. You mentioned in your self-review about routes and I think starting with routes would actually help you here, as it would help you separate concerns right from the start - you can then think of your code as sort of a pipeline where the various segments are:

- Parse the route
- Match the route to a controller method
- Build the model
- Inject the model into the view
- Render the view into the page

Each step is a different concern and can be handled by a different class. If you try and think about how to do each one on it's own it should then be clearer how to build an effective pipeline.

js/models/article.js
```diff
@@ -0,0 +1,35 @@
...
+    }
+
+    return {
+      showHeadline: showHeadline,
+      showSummary: showSummary,
+      showUrlToFullStory: showUrlToFullStory,
+      featureImage: featureImage
+    };
+
+    var headline, url, image, summary
```
mdsomerfield: I'd be interested to see the article you were reading on variable hoisting. I think there's a difference between variable hoisting as something which javascript does, and the pattern in which you have written this code to take advantage of that. In the case of the latter, I feel a bit uneasy about it and it's not clear to me what benefits it has.

PR #17, opened 2017-11-11
guardian's politics feed

mormolis:


js/article-view.js
```diff
@@ -0,0 +1,38 @@
+"use strict";
```
mdsomerfield: For view logic, the concept of 'self-documenting code' is usually a bit different. Which is easier to read:

```
function view(title) {
  return "<div class='my-view'><h1>" + title + "</h1></div>"
}
```

or

```
function getTitle(title) {
  return "<h1>" + title + "</h1>";
}

function getView() {
  var view = "<div class='my-view'>" + getTitle(myTitle) + "</div>";
}
```
Essentially, when you are writing a view, the view is it's own documentation, so I tend to think it's cleaner to keep the chunks larger, and only break out meaningful, reusable components (e.g. a menu or a form) into subviews

js/json-parser.js
```diff
@@ -0,0 +1,58 @@
+"use strict";
```
mdsomerfield: This feels like an over-complication. Why do this:

```
_url[i] = out.response.results[i].webUrl;
_title[i] = out.response.results[i].webTitle;
_body[i] = out.response.results[i].fields.body;
_thumbnail[i] = out.response.results[i].fields.thumbnail;
```
And elsewhere this:
```
article = new Article();
article.setTitle(titles[i]);
article.setBody(bodies[i]);
article.setThumbnail(thumbnails[i]);
article.setUrl(urls[i]);
```
When you could just do this:
```
var article = new Article(results[i].webUrl, results[i].webTitle, results[i].fields.body, results[i].fields.thumbnail);
```

js/json-parser.js
```diff
@@ -0,0 +1,58 @@
...
+      for (var i = 0; i < out.response.results.length; i++) {
+        _url[i] = out.response.results[i].webUrl;
+        _title[i] = out.response.results[i].webTitle;
+        _body[i] = out.response.results[i].fields.body;
+        _thumbnail[i] = out.response.results[i].fields.thumbnail;
+      }
+    }
+
+    var parseJsonFromUrl = function (url, callback) {
+      fetch(url).then(res => res.json()).then((out) => {
```
mdsomerfield: I'm not sure why you have both a parseJsonFromUrl and a parseJsonFromVar, but I would expect that parsing and fetching should be two different responsibilities handled by two different classes.

spec/article-list-model-spec.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+showOff("ArticleList", function(){
+  var articles = new ArticleList();
+  var articleObj = "article stub";
+
+  theThingy(".addArticle adds an article to the list, returnArticleByIndex returns that article stored in 0 position", function(){
```
mdsomerfield: So, I'm deliberately not reviewing the test framework, but....
.
.
.
.
theThingy????

spec/article-model-spec.js
```diff
@@ -0,0 +1,35 @@
+"use strict";
```
mdsomerfield: I generally don't bother testing getters and setters - unless they have side effects (e.g. setting a value results in an event of some sort) they are just extensions of state rather than representing any real logic. We can test them more effectively by testing the code which calls/uses them.

js/index.js
```diff
@@ -0,0 +1,56 @@
...
+    articles.addArticle(article);
+  }
+  console.log(articles.getArticleByIndex(0).getBody());
+  var articlesView = new ArticleListView(articles);
+  articlesView.renderPage();
+  
+  
+  
+  function displayArticleListener() {
+    window.addEventListener("hashchange", displayClickedArticle);
```
mdsomerfield: This feels like the responsibility of a router

js/index.js
```diff
@@ -0,0 +1,56 @@
...
+var executeAfterFetchingData = function () {
+  var titles = guardianParser.getTitles();
+  var urls = guardianParser.getUrls();
+  var bodies = guardianParser.getBodies();
+  var thumbnails = guardianParser.getThumbnails();
+  
+  var articles = new ArticleList();
+  var article;
+  
+  for (var i = 0; i < titles.length; i++) {
```
mdsomerfield: I've commented on this elsewhere, but it feels like this code could be removed completed and the articles generated by the parser

js/index.js
```diff
@@ -0,0 +1,56 @@
...
+    
+    var clickedArticle = articles.getArticleByIndex(articleId);
+    var singleArticleView = new ArticleView(clickedArticle, articleId);
+    console.log(singleArticleView);
+    
+    document
+    .getElementById("description")
+    .innerHTML = singleArticleView.renderWholeArticle();
+    var link = document.getElementById("title" + articleId);
+    link.click();
```
mdsomerfield: I'm not sure what the intent is here

js/index.js
```diff
@@ -0,0 +1,56 @@
...
+  articlesView.renderPage();
+  
+  
+  
+  function displayArticleListener() {
+    window.addEventListener("hashchange", displayClickedArticle);
+  };
+  
+  function displayClickedArticle() {
+    var articleId = window.location.hash.split("#")[1];
```
mdsomerfield: This function feels like it belongs in a controller

js/index.js
```diff
@@ -0,0 +1,56 @@
+"use strict";
```
mdsomerfield: I think you've treated this class as if it's a container for your application logic - I think you should think of it as the class which knows where the bits of your application logic are. So rather than trying to control all of the view and interactions, you're just initialising the different components of your application. That will help you separate concerns I think.

PR #18, opened 2017-11-12
Stay classy San Diego

edlowther:


app/controllers/requests.js
```diff
@@ -0,0 +1,33 @@
+(function(exports) {
+  'use strict';
+
+  exports.RequestController = function(promoListModel, promoListView, RequestClass) {
+    var _promoListModel = promoListModel;
+    var _promoListView = promoListView;
```
mdsomerfield: I don't feel like these two lines add anything - you could just use the injected variables promoListModel and promoListView.

app/controllers/requests.js
```diff
@@ -0,0 +1,33 @@
...
+    var _listController = new ListController(_promoListView);
```
mdsomerfield: Could you inject this?

app/controllers/requests.js
```diff
@@ -0,0 +1,33 @@
...
+      var _httpRequest = new RequestClass();
+      _httpRequest.onreadystatechange = function() {
+        if (_httpRequest.readyState === RequestClass.DONE) {
+          if (_httpRequest.status === 200) {
+            var results = JSON.parse(_httpRequest.responseText).response.results;
+            updatePromoListModel(results);
+          }
+        }
+      };
+      _httpRequest.open("GET", "http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/politics?show-fields=all", true);
```
mdsomerfield: Try to avoid hard-coding the URL - it means the code is harder to re-use. I realise there's only one API in this app, but good practice would to be to have a generate 'doRequest' method that accepted the verb (GET, POST, etc), the url, and any post data, and then have a separate class responsible for defining the API.

app/models/ArticleModel.js
```diff
@@ -0,0 +1,47 @@
+(function(exports) {
+
+  var articleCounter = 0;
+
+  exports.ArticleModel = function(headline, url, body, main) {
+    var _id = articleCounter;
+    articleCounter++;
+
+    var storyBody = document.createElement("div");
```
mdsomerfield: The model feels like the wrong place to be building the DOM. Your model should be responsible for 'what' (i.e. the information to be injected into the view) and your view should know 'how' (i.e. how to display that information).

I note that you've commented on this in your self review, so a bit more info on how that might look:

```
function render(model) {
   var html = "<div class='some-class'><h1 class='title'>{title}</h1></div>"
   var htmlWithModelData = html.replace('{title}', model.title)
   document.getElementById("app").innerHTML= htmlWithModelData;
}
```
EDIT: just seen that you've adopted more of this approach in your view

spec/ControllerSpec.js
```diff
@@ -0,0 +1,116 @@
...
+    return mockRequestInstance;
+  };
+
+  MockRequestClass.DONE = 'yep';
+
+  var controller = new Controller(MockPromoListModel, MockPromoListView, MockRequestClass);
+
+  (function() {
+    console.log(" it delivers an instance of PromoListModel to PromoListView on instantiation");
+    assert.isTrue(mockPromoListView.promoListModel === mockPromoListModel);
```
mdsomerfield: Nitpicking, and not sure if this is within the bounds of the test framework you wrote, but generally assert.areEqual communicates the intent better when comparing two things.

spec/ControllerSpec.js
```diff
@@ -0,0 +1,116 @@
...
+    });
+  })();
+
+  (function() {
+    console.log(" it opens and sends an asynchronous get request to the Guardian news api");
+    assert.isTrue(mockRequestInstance.openCalledCount === 1);
+    assert.isTrue(mockRequestInstance.methodCalled === "GET");
+    assert.isTrue(mockRequestInstance.asynchronousBoolean === true);
+    assert.isTrue(mockRequestInstance.location === "http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/politics?show-fields=all");
+    assert.isTrue(mockRequestInstance.sendCalledCount === 1);
```
mdsomerfield: Do we need all of these assertions? How many of them are needed to test the behaviour, and how many of them are testing your mock?

PR #27, opened 2017-11-14
Incomplete Challenge 

TomSpencerLondon:
I would like to improve my use and understanding of pull requests.


src/stories.js
```diff
@@ -0,0 +1,21 @@
+"use strict";
+
+;(function(exports) {
+   function Stories() {
+    this._storyObjects = [];
+    // this._guardianUrl = "http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/?q=uk&show-fields=body";
+    this._guardianUrl = "https://content.guardianapis.com/search?show-fields=body&api-key=301c8978-6b59-4980-9593-5c8d085394e7";
```
mdsomerfield: This kind of thing usually belongs in a settings class

src/stories.js
```diff
@@ -0,0 +1,21 @@
...
+  };
+
+  Stories.prototype.storiesList = function() {
+    var results = guardianAPI(this._guardianUrl);
+    for(var i = 0; i < results.length; i++){
+      var story = new Story(results[i]);
+      this._storyObjects.push(story);
```
mdsomerfield: I think _storyObjects could be just a local variable - making it a property of Stories makes me think 'oh, what else are we doing with it?'

src/index.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
+
+;(function() {
+  var stories = new Stories();
+  var storiesList;
+  var aylienUrl = "https://api.aylien.com/api/v1/summarize?url=";
+
+
+  window.addEventListener("load", function(){
+    storiesList = downloadStories();
```
mdsomerfield: This feels like quite a lot to put in an anonymous function. Usually, I'd want to a call back to do one thing - if it's a simple thing like 'make this button green' then it's find for the code just to live in the anonymous function. 

In this case though it seems like what we're doing is 'render the page', so we should delegate all of that function to a class, e.g.`window.addEventListener("load", app.render)` (where app is an instance of ab Application class and render is a method responsible for building the page)

src/api.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+function aylienAPI (url, resultsCallback, errorCallback) {
+  var aylienIDHeaderName = "X-AYLIEN-TextAPI-Application-ID"
+  var aylienID = "c11989db";
+  var aylienKeyHeaderName = "X-AYLIEN-TextAPI-Application-Key"
+  var aylienKey = "cac51db271ef8cf0ad06f08e10c4d03c";
+
+  var xhr = new XMLHttpRequest();
+  xhr.onreadystatechange = function() {
```
mdsomerfield: It feels like there's some repetition here between lines 9-18 and 31-40. Could we refactor into a common method?

src/index.js
```diff
@@ -0,0 +1,47 @@
...
+    for (var i=0; i<headlines.length; i++){
+      addHeadlineClickListener(headline,i);
+    }
+  };
+
+  function addHeadlineClickListener(headline, i) {
+    headline.addEventListener('click', function(even) {
+    var elem = document.getElementById('news_summary');
+    var fullAylienUrl = aylienUrl + storiesList[i].getWebUrl();
+    var aylienResult = aylienApi(fullAylienUrl, function(result){
```
mdsomerfield: Generally avoid using anonymous functions where you could use a named method - it makes the app easier to read and (sometimes) easier to test.

src/index.js
```diff
@@ -0,0 +1,47 @@
+"use strict";
```
mdsomerfield: This file seems to be doing 3-4 different things - it's defining the various bits of the app, it's building the view, it's defining some behaviour to the view, and it's then applying the view to the page. 

In your self-review you mentioned: "I have been slow in implementing the code for calling the code onto the html page and would like to focus on building an effective means of doing this." I think that fact that you haven't broken it down a bit more has probably contributed to this - an approach you might try in future:

1) Write the HTML, with dummy data and no behaviour
2) Write some code for adding the code to the page
3) Extract the data into a model - e.g. title, content, image url
4) Write some code to bind data to the view
5) Extract behaviour into your model e.g. callbacks
6) Write some code to bind callbacks to the view

Approaching it in this fashion should help you separate concerns, focus on one thing at a time, and maintain a visual measure of progress.

PR #34, opened 2017-12-10
News summary challenge submission

gabrielabud:


spec/articleListSpec.js
```diff
@@ -0,0 +1,16 @@
...
+describe('ArticleList', function() {
+  var articleList;
+
+  beforeEach(function(){
+    articleList = new articleListModule.ArticleList();
+  });
+
+  describe('#articles', function(){
+    it('has an empty array', function(){
+        expect(articleList.articles()).toBeEmpty();
```
mdsomerfield: State not behaviour!

As an exercise, try laying your tests out in the format:

```
// Arrange
... prepare your test, e.g. create the test instance, mock dependencies etc

// Act
... perform the action that you are trying to test, which should be a method on the instance being tested

// Assert
... your expect(...) statement(s)

```
If you lay your tests out like this and you find that (a) there is no code in the 'Act' section or (b) your test would pass if you commented out the 'Act' section then the chances are that you are testing state rather than behaviour.

This applies to some other tests as well. I dropped a link to a short article into the October channel as this is something that I see in quite a lot of code reviews.

src/app.js
```diff
@@ -0,0 +1,28 @@
+var allnews = new ArticleList()
+
+fetch("http://content.guardianapis.com/search?show-fields=all&q=uk&page-size=12&api-key=fc20c2bc-137d-4f7b-a6c3-3c4efb09563d")
```
mdsomerfield: Ooh, I didn't know 'fetch' was a thing! Looks like it's only been around for a few months, I just checked the API docs (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Specifications) and it looks like it's not supported by IE (something to beware of with newer Javascript features) but full marks for using it and also for making use of promises.

src/app.js
```diff
@@ -0,0 +1,28 @@
...
+
+    buildGrid(allnews._articles);
+
+    var hash = getIDFromUrl();
+    if(hash) {
+        viewSingleArticle(allnews, hash);
+    }
+
+    hashChangeEventListener(allnews);
+  })
```
mdsomerfield: It feels like you are doing a lot in this file - first you fetch the data, then you create a list of models, then you build a grid, then you check the hash, then you add the event listener. It feels like there are a lot of responsibilities here. 

Fetching the data feels like it belongs in an API service class, building the models and returning a view feels like it is part of the controller, interpreting the hash belongs in a router. Setting up the hash event listener and building the controllers (i.e. application setup) feel like the only things that we should really have in an app.js in this example.

src/articleList.js
```diff
@@ -0,0 +1,31 @@
...
+    article._id = idnum;
+    return this._articles;
+  },
+
+  articles: function() {
+    return this._articles;
+  },
+
+  getArticleById: function(id) {
+    for(var i = 0; i < this._articles.length; i++) {
```
mdsomerfield: You could use _articles.find() here - see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/prototype for a full list of collection methods

src/articleView.js
```diff
@@ -0,0 +1,55 @@
+function getIDFromUrl() {
```
mdsomerfield: There's not a lot of OO going on here - it's just a collection of functions. In such a small application that probably doesn't feel too bad, but in a larger application it would be pretty hard for another developer to come in and work out where all of these methods fitted into the larger application flow, and you'll also quickly start to feel pain when testing because it's difficult to isolated dependencies.

src/buildGrid.js
```diff
@@ -0,0 +1,22 @@
+function buildGrid(articles) {
+  var root = document.getElementById("root");
+  var thumbs = document.createElement("div");
+  root.innerHTML = "<h1>Ame News</h1>";
+
+  root.appendChild(thumbs);
+  thumbs.setAttribute("class","thumbs");
+  thumbs.setAttribute("id","thumbs");
+
+  articles.forEach(function(item, index) {
```
mdsomerfield: Feels like this could be abstracted into a buildArticle method and then you could just do:
```
articles.forEach(buildArticle);
```
And you would have documented the intent more clearly.

PR #44, opened 2017-12-11
news-summary-challenge

majdeddine:


src/article-controller.js
```diff
@@ -0,0 +1,72 @@
...
+      document
+      .getElementById('app')
+      .innerHTML = self._articleListView.toHtml();
+      self.hashChange();
+    },
+
+    hashChange: function() {
+      var self = this;
+      var content = this._articleListView._list;
+      function showArticleForCurrentPage() {
```
mdsomerfield: The hash change event looks like any hash change assumes that you are navigating to the single article view. What if somebody navigates back to the main article list?

A cleaner implementation uses a route table approach, e.g.

```
var routes = [{
  route: /articles/, method: showArticleList,
  route: /articles\/(\d+)/, method: showArticle,
}]

var method = routes.find((route) => {
  return route.test(window.location);
});

method();
```

This method uses regular expressions to match the url (see https://eloquentjavascript.net/09_regexp.html if you are not familiar with these, I realise we haven't covered them on the course). You don't have to use regular expressions, but the point is that you should have a list of routes and a list of controller methods that match them, and then execute the method based on which route matches.

NB I didn't think very long about the code above so it may have a bug or two...

src/article-list-view.js
```diff
@@ -0,0 +1,42 @@
...
+    this._list.forEach(function(article){self._result+= renderToHtml(article)})
+
+  }
+
+  ArticleListView.prototype.toHtml = function(){
+    return this._result;
+  }
+
+  ArticleListView.prototype.addArticle = function(data, body){
+    this._articleList.add(data, body);
```
mdsomerfield: It feels odd that you need to pass an ArticleModel into the view in order to add it to the ArticleListModel - this feels like tight coupling. I would expect you to be able to add the ArticleModel directly to the ArticleViewModel and the for the ArticleListView not to need this method or to know anything about how the ArticleListModel is built.

PR #53, opened 2018-01-15
news-summary-challenge

Xin00163:


src/list-view.js
```diff
@@ -0,0 +1,16 @@
+(function(exports){
+  var View = function(list){
+    this.list = list;
+  }
+
+  View.prototype.displayHeadlines = function() {
+    return "<ul>" +
+
+    this.list.returnAllStories().map( story => "<li id=\"" + story.id + "\"><img src=\"" + story.thumbnail + "\"></br></br><a href=\"#\" onclick=\"readStory(" + story.id + ")\">" + story.getTitle() + "</a></li></br></br>").join("") + "</ul>"
```
mdsomerfield: I like this approach - but be careful about putting it all on one line. It's more readable if you break it up. Also, remember you can use double and single quotes in javascript, so then you don't need so many escape characters (which also make it hard to read). e.g something like:

```
this.list.returnAllStories().map((story) => {
   "<li id='" + story.id + "'>" +
      "<img src='" + story.thumbnail + "'>" +
      "</br></br>" +
      "<a href='#' onclick='readStory(" + story.id + ")'>" + story.getTitle() + "</a>"
   "</li>").join("");
});
```
   
I can look at that and it's more obvious what the output will be as it's formatted more like I would format raw HTML.

mdsomerfield (commented):
Hi Xin,

This looks pretty good. A couple of comments from your self-review: 

*Q: What do you think was hard or confusing? A: How to handle API request*

I left a comment and linked to a short snippet of code regards the API request. It looks like you've successfully connected to the API and the code you've written looks like more or the less the code I would expect; I would suggest however that the concerns could be separated and made more DRY. Let me know if the example I have linked to helps or if you want to explore this further.  

*Q: What do you want to improve on: Figuring out what models and views do I need. For instance, maybe the list model will take care of requested data then the story model will be redundant. Instead, I could have a summary model and summary view.

I actually think your model is fine - and also the way you've integrated your controller, view and model looks like a solid interpretation of the MVC pattern. So I actually don't have a lot of feedback here, other than well done.

PR #61, opened 2018-02-18
Pull request LG

Leigan0:


src/models/ArticleSummaryRequest.js
```diff
@@ -0,0 +1,27 @@
...
+    this.articleUrl = url
+  }
+
+  ArticleSummaryRequest.prototype.requestSummary = function(callback,HttpRequest = new XMLHttpRequest){
+    var xhttp = HttpRequest
+    var self = this
+    xhttp.onreadystatechange = function(){
+      if (xhttp.readyState == 4 && xhttp.status === 200) {
+          var data = JSON.parse(xhttp.responseText)
+          self.summary = data.sentences.join()
```
mdsomerfield: It feels like this is a separate responsibility. We have two things going on here - sending the request and acting on the response. I would suggest that you just pass the data to the callback, e.g:

```
var data = JSON.parse(xhttp.responseText);
callback(data);
```

It's the responsibility of the callback to know what to do with that data. This is a pretty standard pattern for API calls in javascript.

src/controller.js
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  function onLoad(){
+    newslog = new NewsLog()
+    newsview = new NewsHeadlinesView(newsLog())
+    newslog.getArticles(newsView().displayHeadLines)
+    showSelectedSummary()
+  }
+
+  exports.onLoad = onLoad
```
mdsomerfield: It feels like what you really want to export here is a Controller class - a lot of the logic could be a in Controller but some of it belongs elsewhere. 

A good pattern is to have an app.js file which handles application setup. Then rather than <body onload="onload()"></body>, I would remove the onload event from the HTML and in your app.js have some code that looks a bit like:

```
function onLoad(){
  newslog = new NewsLog()
  newsview = new NewsHeadlinesView(newsLog)
  controller = new Controller(newsview);
  controller.initialise();
}

if (document.readyState === 'complete') {
  onload();
} else {
  document.onload = onload;
}
```
Your controller can then contain other initialisation code relating to the different views (e.g. showSelectedSummary and newslog.getArticles etc). You're effectively distinguishing here between application *setup* (create all your objects and inject relevant dependencies) and then initialising the application *flow* (getting data, rending html etc).

src/controller.js
```diff
@@ -0,0 +1,47 @@
...
+
+  function requestSummary(index){
+    var articleUrl = newslog.articles[index].url
+    summaryrequest = new ArticleSummaryRequest(articleUrl)
+    summaryrequest.requestSummary(newsView().displayArticleSummary)
+  }
+
+  function onLoad(){
+    newslog = new NewsLog()
+    newsview = new NewsHeadlinesView(newsLog())
```
mdsomerfield: I think there is some confusion over the use of brackets - when you add () it executes the object as if it were a function. But newslog here looks like an object not a function. 

You mention in your self-review that you are not sure about callbacks yet as well - this is something lots of people struggle with. I've yet to find a really good resource that demystifies it. Maybe took a look at https://gist.github.com/amysimmons/3d228a9a57e30ec13ab1 (and let me know if that helps)

test/articleSummaryRequestTest.js
```diff
@@ -0,0 +1,46 @@
...
+})
+
+it('stores url argument passed at instantiation', function (){
+  var summaryRequest = new ArticleSummaryRequest('https://www.theguardian.com/artanddesign/2018/feb/18/leslie-fairweather-obituary')
+  return assert.isTrue(summaryRequest.articleUrl, 'https://www.theguardian.com/artanddesign/2018/feb/18/leslie-fairweather-obituary')
+})
+
+it('stores sentences returns from api apiRequestResponse',function(){
+  var summaryRequest = new ArticleSummaryRequest();
+  summaryRequest.requestSummary(callback,successfulHttpRequest)
```
mdsomerfield: You mentioned in your self review that you were struggling to isolate your API calls. I've left some feedback in a comment above regards injecting in the constructor - I think this will help. If you do that, you can then create a stub function:

```
function mockHttpRequest() {
  ...
}
var summaryRequest = new ArticleSummaryRequest(mockHttpRequest); 
```

So very similar to what you are doing here - I think you nearly have it - but instantiating the object with the dependency rather than passing it into the method and making it a constructor *function* rather than an instance.

PR #135, opened 2018-11-18
both APIs used with own key, CSS styling (Hacker News clone), images, no animations, mostly test-driven

hnryjmes:


imogenmisso (commented):
You have managed to successfully use ES6 in this challenge! You are clearly confident learning new things by yourself.

Looking from your commits, I can see you have added the image feature quite late. Your code is clearly easy to change as you were able to add this without any issues.

Nice use of setTimeout to create asynchronous code and ensure your tests pass! Again this is something we have not covered during the week - it is impressive you managed to learn so much during the weekend!

PR #172, opened 2019-02-05
bit late 

GLMill:


public/scripts/base_display_handler.js
```diff
@@ -0,0 +1,61 @@
+
+var base_display_handler = function(handler){
+
+    function convert_all_to_thumb(array = news_source_articles, target_node){
```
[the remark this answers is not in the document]
alicelieutier: re: naming. array is not a meaningful name for an array. what does it contain?

public/scripts/base_display_handler.js
```diff
@@ -0,0 +1,61 @@
...
+        const parent_node = document.getElementById(target_node);
+        for(let event of array){
+            let thumb = create_thumbnail(event);
+            parent_node.appendChild(thumb);
+        }
+    };
+
+    function make_text_element(element, string){
+        let _ = document.createElement(element);
```
alicelieutier: naming again. `_` should only be used for variables that will never be used.

public/scripts/base_display_handler.js
```diff
@@ -0,0 +1,61 @@
...
+    };
+
+    return {
+        convert_all_to_thumb : convert_all_to_thumb,
+        make_image : make_image,
+        make_article_element: make_article_element,
+        make_link : make_link,
+        make_text_element: make_text_element,
+        create_thumbnail: create_thumbnail
+    }
```
alicelieutier: in ES6, if the value and the key have the same name, you don't have to write them twice:
`{a: a, b: b, c: d}` can be written `{a, b, c: d}`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#New_notations_in_ECMAScript_2015

alicelieutier: Do you need to export all these methods? Should some of them stay "private" ?

public/scripts/fetch_request.js
```diff
@@ -0,0 +1,20 @@
...
+            return response.json();
+         })
+         .then(function(myJson){
+            fetch_call_back(myJson)
+         })
+   })();
+   /* is this call back doing to much? i just liked the idea of being able to give
+      the fetche request an api address, how to display events and where - ifelt it would allow greater
+      felxibility... ?
+    */
```
alicelieutier: Definitely not doing too much. It's normal to have most of your app's code triggered through callbacks. What is important is that the code is neatly organised per concern.

public/scripts/guardian_display_handler.js
```diff
@@ -0,0 +1,41 @@
+var guardian_display_handler = (function(response, target_node){
+
+    const base = base_display_handler;
+
+  
+    /* make elements creates the paragraph elements */
+    function make_elements(obj){
```
alicelieutier: This method is only called when you are calling create_thumbnail. Basically, you need that method to create arguments in the right format for create_thumbnail.
A better option would be to pass `obj` to `create_thumbnail` directly, and then call `make_elements` on `obj` inside `create_thumbnail`

public/scripts/guardian_display_handler.js
```diff
@@ -0,0 +1,41 @@
...
+                    base.make_link( obj.webUrl, obj.id)
+                ]
+
+    if(obj.fields){
+        elements.push(base.make_image(obj.fields.thumbnail),)
+    }
+    return elements
+    }
+    
+    /* both functions below seem repetitive and could be seperated out? */
```
alicelieutier: not sure which functions you mean

PR #245, opened 2020-01-12
Weekend 7 News Summary Challenge

AndreaDiotallevi:


spec/articleListControllerSpec.js
```diff
@@ -0,0 +1,24 @@
...
+  var articleListViewDouble = new ArticleListViewDouble(articleListDouble)
+  var guardianApiRequestDouble = new XMLHttpRequestDouble()
+  var articleListController = new ArticleListController(articleListDouble, articleListViewDouble, guardianApiRequestDouble)
+  articleListController.fetchArticleList()
+  assert.isTrue(articleListController._articleList.articles().length === 1)
+  assert.isTrue(articleListController._articleList.addArticleCallCount === 1)
+  assert.isTrue(articleListController._guardianApiRequest.openCallCount === 1)
+  assert.isTrue(articleListController._guardianApiRequest.sendCallCount === 1)
+  // assert.isTrue(articleListController._request.onloadCallCount === 1)
+})()
```
alicelieutier: This whole test is testing state more than behaviour.
Just the fact that you have to access a private variable `articleListController._articleList` in your test tells me that.
The next test is slightly better, but it mat be event better to separate the logic (creating the HTML for the list) and testing that against the expected string, and the display (adding that string to your html document).

In terms of mocking, you don't need to check that methods have been called on your double (using your "count" methods) to write good behaviour tests, unless the only behaviour from the method is calling that other method).

alicelieutier: If you used fetch, you could mock like this:
```js
let savedFetch = window.fetch
window.fetch = function(){ /* stubbing behaviour here */ }
// your test here
window.fetch = savedFetch // make sure you put the things back where they belong.
```

alicelieutier: You could stub XMLHttpRequest in the same way if you wanted.

PR #268, opened 2020-04-06
code review

lukedonov:


src/storyController.js
```diff
@@ -0,0 +1,19 @@
+(function(exports) { 
+  
+  function StoryController(list) {
+    this.storyList = list;
+    this.storyListView = new StoryListView(list);
+  }
+
+  StoryController.prototype.outputToHtml = function() {
```
lukedonov, line 8: @jamesjoshuahill This is the method that doesn't seem to be working. If I run controller.storyListView.viewHeadlines() it outputs what I want so I'm not too sure why this.storyListView doesn't do the same.

jamesjoshuahill: Hi @lukedonov I was able to recreate this issue. When I first load the page I see a single bullet point with no text. When I run `controller.outputToHtml()` in the JavaScript console, then I see the list of headlines. So something changed between the page loading and running the command on the console 🤔 

In the `list.getStories()` function, **when** is the list of stories populated?

PR #273, opened 2020-05-03
Guardian newsfeed

Edward-Phillips:
Asynchronous js really threw me for a loop, I had it right initially but then became convinced I didn't as I didn't realise you needed to call .response from the object that was returned.

I thinik I've covered all 6 user stories, if you want to have a look at the site without git cloning and running locally have a look at http://guardian-top-10-feed.surge.sh/

Areas to improve:
styling: it looks nice but isn't very original, largely pilfered from https://taniarascia.github.io/sandbox/ghibli/ 
I'd like the headlines to be centered below the images, which I'm just blanking on that at the moment.

The page is being populated in a big chunky function that could probably be slimmed down a bit but it took me so long to work out how to use the api result I've kind of run out of steam in terms of refactoring - still open to any suggestions on refactoring though!


.gitignore
```diff
@@ -0,0 +1,4 @@
+#keep it secret, keep it safe.
```
hturnbull93, line 1: I'm not trying to rob you, I'm trying to help you.

index.html
```diff
@@ -0,0 +1,19 @@
...
+  <link rel="stylesheet" href="style.css">
+  <script src="./public/interface.js"></script>
+  <script src='./api-stuff.js'></script>
+  <script src='./public/newsitem.js'></script>
+  <script src='./public/newsreel.js'></script>
+  <title>Guardian Api Newsfeed</title>
+</head>
+<body>
+  <h1>Guardian Newsfeed</h1>
+  <div id='root'>
```
hturnbull93, line 15: what is this, react?

public/interface.js
```diff
@@ -0,0 +1,26 @@
...
+            text = article.getbodyTextHTML(),
+            thumbnail = article.getThumbnailHTML();
+            container.id = index;
+            container.classList.add('article-shell')
+            shell = document.getElementById('news-spot');
+            shell.appendChild(container);
+            container.appendChild(thumbnail);
+            container.appendChild(title);
+            container.appendChild(text);
+          }
```
hturnbull93, line 23: Could this be extracted to a method?

public/newsitem.js
```diff
@@ -0,0 +1,32 @@
...
+  this.shortUrl = apiResponseElement.shortUrl;
+  this.thumbnail = apiResponseElement.thumbnail;
+  this.id = articleID
+}
+
+newsItem.prototype.getHeadlineHTML= function() {
+  var headline = document.createElement('a')
+  headline.href = this.shortUrl;
+  headline.classList.add('headline');
+  headline.id = `Article #${this.id} headline`;
```
hturnbull93, line 13: this will result in three ids, Article, 4, and headline. Article and headline are then not unique. It should probably be something like:
```
headline.id = `Article-#${this.id}-headline`;
```
If that's what you're going for?

public/newsitem.js
```diff
@@ -0,0 +1,32 @@
...
+  headline.textContent = this.headline;
+  return headline;
+};
```
hturnbull93, line 16: For HTML semantics you could have a h2 or h3 wrapped in the a? At the moment its semantically no different from the paragraph text below

public/newsitem.js
```diff
@@ -0,0 +1,32 @@
...
+  bodyText.id = `Article #${this.id} bodytext`;
+  bodyText.classList.add('body')
+  bodyText.textContent = this.bodyText.substr(0,200) + '...';
+  return bodyText;
+};
+
+newsItem.prototype.getThumbnailHTML = function() {
+  var thumbnail = document.createElement('img');
+  thumbnail.classList.add('thumbnail')
+  thumbnail.id = `Article #${this.id} image`;
```
hturnbull93, line 29: same comment as before

public/newsitem.js
```diff
@@ -0,0 +1,32 @@
...
+  headline.href = this.shortUrl;
+  headline.classList.add('headline');
+  headline.id = `Article #${this.id} headline`;
+  headline.textContent = this.headline;
+  return headline;
+};
+
+newsItem.prototype.getbodyTextHTML= function() {
+  var bodyText = document.createElement('p');
+  bodyText.id = `Article #${this.id} bodytext`;
```
hturnbull93, line 20: same comment as before

PR #276, opened 2020-05-04
Alexa's News

alexakearns:


api.js
```diff
@@ -0,0 +1,24 @@
+var url = 'https://content.guardianapis.com/search?api-key=fed10329-e5b5-4ed3-afda-6ac7a6d00245'
```
tatsiana-makers, line 1: Under the "Guardian API example" section of the challenge README, there is another API that you should use for this challenge. 

It's considered bad practice to store API keys in the frontend - can you think of a reason why this might be?

api.js
```diff
@@ -0,0 +1,24 @@
...
+var newsdata = []
+var apiKey = 'fed10329-e5b5-4ed3-afda-6ac7a6d00245'
+
+function getApi() {
+    fetch(url)
```
tatsiana-makers, line 6: Nice use of the fetch API!

src/article.js
```diff
@@ -0,0 +1,13 @@
...
+function Article(id, type, sectionId, sectionName, webPublicationDate, webTitle, webUrl, apiUrl) {
+    this.id = id
+    this.type = type 
+    this.sectionId = sectionId
+    this.sectionName =sectionName
+    this.webPublicationDate = webPublicationDate
+    this.webTitle = webTitle
+    this.webUrl = webUrl
+    this.apiUrl = apiUrl
+
```
tatsiana-makers, line 12: Currently, the `Article` doesn't own its displaying. If you think back to one of the demos Alice did for the Notes app, we had the Notes component owning how its displaying itself.

The logic currently lives in https://github.com/makersacademy/news-summary-challenge/pull/276/files#diff-ec681c37e2090e8a8320da5dab21f4b7R16

but we could move this logic into a method attached to the Article component

views/interface.js
```diff
@@ -0,0 +1,53 @@
...
+    title.innerHTML = article['webTitle']
+    newsArticle.appendChild(newspiece)
+    newspiece.appendChild(title)
+    console.log("working!")
+
+    }
+    console.log("Hello")
+}
+
+updateNews()
```
tatsiana-makers, line 49: We call `updateNews` both here and within `getApi`. Do we need both?

views/interface.js
```diff
@@ -0,0 +1,53 @@
+document.addEventListener('DOMContentLoaded', function(event) {
+    console.log("DOM content loaded")
+
+    getApi()
+    console.log(newsdata)
+});
+
+var newsArray = []
```
tatsiana-makers, line 8: Is it worth about pulling this out into it's own component, just as we did for `Article`?

PR #281, opened 2020-05-31
Zsofi's news summary challenge

ZsofiaS:


public/index.html
```diff
@@ -0,0 +1,25 @@
...
+    <meta charset="utf-8">
+    <link href="https://fonts.googleapis.com/css2?family=Lato&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="css/style.css">
+    <title>News summary</title>
+  </head>
+  <body>
+    <h1>News headlines</h1>
+    <div id="app">hello</div>
+
+    <script type="text/javascript" src="js/src/articleModel.js"></script>
```
will-head, line 13: Great separation into classes - really clear what each is doing!

public/index.html
```diff
@@ -0,0 +1,25 @@
...
+    <script type="text/javascript" src="js/src/articleListModel.js"></script>
+    <script type="text/javascript" src="js/src/articleListView.js"></script>
+    <script type="text/javascript" src="js/src/articleView.js"></script>
+    <script type="text/javascript" src="js/src/articleController.js">
+    </script>
+    <script type="text/javascript">
+      let articleController = new ArticleController()
```
will-head, line 20: Would it make more sense for ArticleController to use class methods?

public/js/spec/articleListModelSpec.js
```diff
@@ -0,0 +1,40 @@
+(function() {
```
will-head, line 1: Lots of test coverage! 👍

public/js/src/articleController.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  ArticleController.prototype.insertHTML = function() {
+    let app = document.getElementById('app')
+    app.innerHTML = this.articleListView.returnHTMLstring();
+  }
+
+  ArticleController.prototype.showSingleArticle = function() {
+    let self = this;
+    window.addEventListener('hashchange', function() {
```
alicelieutier, line 14: It would make more sense for me to have this as part of your constructor, since this is about initialising the event handlers, and always needs to be done when you are setting up the controller.

public/js/src/articleListModel.js
```diff
@@ -0,0 +1,29 @@
...
+
+  ArticleList.prototype.getData = function() {
+    let self = this
+    fetch('http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/travel')
+      .then(response => response.json())
+      .then(data => {
+        let results = data.response.results;
+        results.forEach((item, i) => {
+          self.add(item.webTitle)
+          self.list[i].url = item.apiUrl
```
alicelieutier, line 22: I would add a class here called article, to get to code that looks like this:
```js
results.forEach((item) => {
          let article = new Article(item.webTitle, item.apiUrl)
          self.add(article)
});
```
Then each article is responsible for its own title and URL, and could also produce its own html.

If you don't want another class, it may still be cleaner to do:
```js
results.forEach((item) => {
          let article = {title: item.webTitle, url: item.apiUrl)
          self.add(article)
});
```

public/js/src/articleListModel.js
```diff
@@ -0,0 +1,29 @@
...
+        });
+        articleController.insertHTML()
```
alicelieutier, line 24: This is not the best place to have that, as it makes you ArticleList and ArticleController classes very coupled. Instead, you could have something like:
in ArticleList:
```js
ArticleList.prototype.getData = function(callback) {
    let self = this
    fetch('http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/travel')
      .then(response => response.json())
      .then(data => {
        let results = data.response.results;
        results.forEach((item, i) => {
          self.add(item.webTitle)
          self.list[i].url = item.apiUrl
        });
      callback()
      })
  }
```

And in articleController's constructor:
```js
let self = this
this.articleList.getData(self.insertHTML)
```

This way you know `insertHTML` will run after the results have come back, but the ArticleList class doesn't need to know about functionalities of the ArticleController.

public/js/src/articleModel.js
```diff
@@ -0,0 +1,12 @@
+(function(exports) {
```
alicelieutier, line 1: So you totally have an article class. You should use this to store information in the article list.

public/js/src/articleView.js
```diff
@@ -0,0 +1,25 @@
...
+        .then(response => response.json())
+        .then(data => {
+          let result = data.response.content;
+          let content = result.fields.body
+          stringHTML = content;
+          document.getElementById('app').innerHTML = stringHTML;
+        })
+    } else {
+      stringHTML = '<ul><li><div>' + this.article.headline + '</div></li></ul>'
+      document.getElementById('app').innerHTML = stringHTML;
```
alicelieutier, line 19: The part of your code that attaches new element to the DOM should be in the controller not the view (to be consistent with what you do for article list). Also because this function is called `returnHTMLstring` so it should just return the string, not also change the DOM.

PR #283, opened 2020-06-01
theyesterday_news

tanilserbes:


public/src/news.js
```diff
@@ -0,0 +1,59 @@
+const app = document.getElementById('root')
+const newlist = document.querySelector('.new-list')
+
+const logo = document.createElement('img')
+logo.src='images/logo.png'
+
+const container = document.createElement('div')
+container.setAttribute('class', 'container')
```
alicelieutier, line 8: It's ok to have the logo and the container be created by the HTML if you know they will always be there.

public/src/news.js
```diff
@@ -0,0 +1,59 @@
...
+
+
+app.appendChild(logo)
+app.appendChild(container)
+
+
+fetch('https://content.guardianapis.com/search?show-blocks=all&api-key=3d1bd118-3d6f-4211-a61e-3d0f966b08bb')
+
+.then(response => {
```
alicelieutier, line 17: There should be no new line between these. You are calling the method `then` on the object returned by the function `fetch`.

public/src/news.js
```diff
@@ -0,0 +1,59 @@
...
+  return response.json()
+  
+})
+.then(data => {
+    data.response.results.forEach(news => {
+     var li = document.createElement('li')
+     var a = document.createElement('a')
+     a.setAttribute('href', onclick=retrieve)
```
alicelieutier, line 25: What does this line do?

public/src/news.js
```diff
@@ -0,0 +1,59 @@
...
+     a.text = "see article"
+     const card = document.createElement('div')
+     card.setAttribute('class', 'card')
+     const h3 = document.createElement('h3')
+     h3.textContent = news.webTitle
+     li.appendChild(a)
+     container.appendChild(card)
+     card.appendChild(h3)
+     card.append(li)
+     window.webUrl=news.webUrl
```
alicelieutier, line 36: Not sure what you are trying to do with this line. Right now, what you are doing is creating a global variable called `webUrl` and writing the piece of news URL in it for each piece of news. Which means in the end you have overwritten it a number of times.

public/src/news.js
```diff
@@ -0,0 +1,59 @@
...
+  
+})
+.then(data => {
+    data.response.results.forEach(news => {
+     var li = document.createElement('li')
+     var a = document.createElement('a')
+     a.setAttribute('href', onclick=retrieve)
+     a.setAttribute('target', '_blank')
+     a.text = "see article"
+     const card = document.createElement('div')
```
alicelieutier, line 28: This all code to create elements should probably be extracted to its own method. Maybe something called `createCard(news)` ?

public/src/news.js
```diff
@@ -0,0 +1,59 @@
...
+function retrieve(e){
+  e.preventDefault()
+
+  var settings = {
+    "url": `https://api.meaningcloud.com/summarization-1.0?key=c855ca1f61d1002436b3f724a17c3a72&url=${window.webUrl}&sentences=2`,
+    "method": "POST",
+    "timeout": 0,
+  };
+
+  $.ajax(settings).done(function (response) {
```
alicelieutier, line 54: Here you use ajax, but above you use fetch. They both do the same thing. It would be better to be consistent and use only one of them.

public/src/news.js
tanilserbes: fixed.

public/src/news.js
tanilserbes: Ekran Resmi 2020-06-02 07 49 29

i deleted li html element now. 
Creating a element in every each news and then when user clicks in the browser, it triggers retrieve function which runs summarized article.

tanilserbes: Problem here is, when i click in anywhere in the browser, it works.(the retrieve function works). It is supposed to work when i only click 'see article' writing.

public/src/news.js
tanilserbes: You are right, it doesn't work properly. 
Here, creating global webUrl and then the retrieve function takes this Url. Otherwise, i can't transfer the url to the outside function... AHA!! or this is the time i am gonna use callback. ?????
Every time it takes same Url and user clicks the see article, it shows same summarized article.

public/src/news.js
tanilserbes: editted.

public/src/news.js
tanilserbes: yes, thanks for mentioning. api provides ajax. 
I couldn't make it with fetch.

tanilserbes: https://www.meaningcloud.com/developer/summarization/dev-tools/1.0

PR #293, opened 2020-06-29
News Summary Week 7 weekend challenge 

ad13380:


index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>News Summary</title>
+
+</head>
+<body onload="new NewsController().renderHeadlines()">
```
tatsiana-makers, line 9: Could we pull this out into an `interface.js`, similar to [the Thermostat challenge](https://github.com/makersacademy/thermostat-example/blob/master/src/interface.js)? 

This could also be a good place to implement the event listeners 🙂

src/article-view.js
```diff
@@ -0,0 +1,16 @@
+(function(exports) {
+  function ArticleView(articleJSON, articleModel) {
```
tatsiana-makers, line 2: Does this need to take both `articleJSON` and `articleModel`? Doesn't the articleModel already contain the json?

src/article-list-view.js
```diff
@@ -0,0 +1,18 @@
+(function(exports) {
+  function ArticleListView(articleList) {
+    this.articleList = articleList;
+  }
+
+  ArticleListView.prototype.returnHTML = function() {
+    return this.articleList.getList().reduce((acc, val) => {
```
tatsiana-makers, line 7: Can we use the logic inside of `ArticleView` to help us render the HTML for each individual article?

src/article-list-model.js
```diff
@@ -0,0 +1,20 @@
+(function(exports) {
```
tatsiana-makers, line 1: Nice use of the constructor/prototype and module patterns!

src/news-controller.js
```diff
@@ -0,0 +1,61 @@
...
+      return article.getId().toString() === this._returnIdFromURL(window.location);
+    })[0];
+  }
+
+  NewsController.prototype._returnIdFromURL = function(location) {
+    return location.hash.split("#")[1].split("/")[1];
+  }
+
+  exports.NewsController = NewsController
+})(this)
```
tatsiana-makers, line 61: Writing tests for this code would be good practice of testing in JS 🙂

PR #301, opened 2020-08-24
news-summary-challenge partial solution

ROCHAAL:


index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
```
tatsiana-makers, line 1: If you open up `index.html` in the browser and open the web console, you'll see a few errors are thrown. The error messages will helpfully point out which line of code the error is coming from 🙂

news-tests.js
```diff
@@ -0,0 +1,16 @@
...
+    function testsNewsSummaryHasHeadlines() {
+        var news = new News("headlines");
+        assert.isTrue(news.headlines === "headlines");
+    }
+
+    function testsNewsCanShowHeadlines(){
+        var news = new News("articles")
+        assert.isTrue(news.showHeadlines() === "articles");
+    }
+
```
tatsiana-makers, line 12: Nice tests (:

index.html
```diff
@@ -0,0 +1,41 @@
...
+        };
+        function showTheGuardianHeadlines() {
+            showHeadlines(getHeadlinesFromUrl(window.location));
+        };
+
+        function getHeadlinesFromUrl(location) {
+            return loction.hash.split("#")[1];
+        };
+
+        function showHeadlines(headlines) {
```
tatsiana-makers, line 34: Cool that you got the text on the page to change!

news.js
```diff
@@ -0,0 +1,10 @@
+(function (exports) { 
+    function News(headlines) {
```
tatsiana-makers, line 2: It doesn't look like we're currently using this `News` component in the `index.html` code. What is the responsibility of this module and how might it eventually fit into the page?

news-views-tests.js
```diff
@@ -0,0 +1,9 @@
+(function (exports){
+
+    function NewsDisplayHeadlines() {
```
tatsiana-makers, line 3: What will the `NewsDisplayHeadlines`'s responsibilities be?

tatsiana-makers (commented):
This is a nice start of the challenge! I've left some comments including some things to think about. 🙂 

In the future, we'll also want to add the API call to actually fetch the data for the page, which we'll use to fill the page with the real articles 🚀

PR #327, opened 2021-01-24
News Summary Challenge (Ara)

aracho1:
- Completed all user stories.

- Hosted on surge http://makers-news-today.surge.sh/

- The Aylien API seems to have a very low limit so I have replaced the summary with full body text when the fetch request fails

- It takes a while for the page to load the articles - wonder if I need to restructure my code to speed up the fetch request/response


scripts/news.js
```diff
@@ -0,0 +1,35 @@
...
+    return `<div class="modal">
+              <span class="btn-close">&times;</span>
+                <div class="modal-content">
+                  ${this.createModalTitle()}
+                  ${this.createThumbnail()} 
+                  ${this.createSummary()}
+                </div>
+            </div>`
+  }
+}
```
PiperS52, line 35: These methods are a really neat way of creating the html

scripts/index.js
```diff
@@ -0,0 +1,52 @@
...
+    let summaryNews = summaryApi + newsUrl;
+    fetch(summaryNews)
+    .then((res) => res.json())
+    .then((data) => {
+      let summaryText = document.getElementById(`summary-${headline.collection.length-1}`)
+      summaryText.innerHTML = data["text"];
+    })
+    }
+  )});
+
```
PiperS52, line 37: Nice work creating the modals

scripts/spec/newsSpec.js
```diff
@@ -0,0 +1,11 @@
...
+  let news = new News();
+  news.title = "today's news";
+  news.url = "http://news.com";
+  news.thumbnail = "http://news.jpg";
+  news.summary = "this is today's news";
+  expect(news.title).toEqual("today's news")
+  expect(news.url).toEqual("http://news.com")
+  expect(news.thumbnail).toEqual("http://news.jpg")
+  expect(news.summary).toEqual("this is today's news")
+})
```
PiperS52, line 11: I wonder whether this could possibly be broken down into multiple it/expects

PiperS52:
This is great - no adverts as well, where can I subscribe!? I love how you even got round to hosting it. Nice work

scripts/date.js
```diff
@@ -0,0 +1,9 @@
+const today = new Date();
+const date = String(today.getDate())
+const month = new Intl.DateTimeFormat('en-US', { month: 'long'}).format(today)
+const day = new Intl.DateTimeFormat('en-US',{ weekday: 'long' }).format(today)
+const year = today.getFullYear();
+
+const displayDate = `${day} ${month} ${date}, ${year}`;
+
+document.getElementById("today-date").innerHTML = `<h3>${displayDate}</h3>`
```
alicelieutier, line 9: This file mixes logic and presentation. Maybe you could wrap the functionality to format today's date into a function, and then in the interface, where you set up the rest of the page, you can replace it.

alicelieutier: That would also avoid declaring all these constants to the global scope.

alicelieutier: So:
```js
function formattedTodaysDate() {
  const today = new Date();
  // ...
  return `${day} ${month} ${date}, ${year}`
}
```

Or maybe even
```js
function formattedDate(date) {
  const month = new Intl.DateTimeFormat('en-US', { month: 'long'}).format(date)
  // ...
  return `${day} ${month} ${monthDay}, ${year}`
}

// then call like this for todays'
formattedDate(new Date())
```

scripts/index.js
```diff
@@ -0,0 +1,52 @@
...
+  const headline = new Headlines();
+  content.forEach(function(news) {
+    const article = new News(news.webTitle, news.webUrl, news.fields.thumbnail, news.fields.bodyText, news.sectionName);
+    headline.add(article)
+    output += `
+      <div class="news-item" id="${headline.collection.length-1}"> 
+        ${article.createThumbnail()} 
+        ${article.createTitle()} 
+        ${article.createModal()}
+      </div>`;
```
alicelieutier, line 21: This is nice, but it could maybe be encapsulated into the headlines class?

alicelieutier: Bhind a `headline.createHeadlines()` maybe?

scripts/index.js
```diff
@@ -0,0 +1,52 @@
...
+  
+    triggerModals();
+    
+    let newsUrl = headline.collection[headline.collection.length - 1].url
+    let summaryNews = summaryApi + newsUrl;
+    fetch(summaryNews)
+    .then((res) => res.json())
+    .then((data) => {
+      let summaryText = document.getElementById(`summary-${headline.collection.length-1}`)
+      summaryText.innerHTML = data["text"];
```
alicelieutier, line 33: You are fetching the summary for all articles when the page is loaded. It would be better IMO to fetch the summary only when a user clicks on a headline. And it would use less of Aylien, for more consistent results.

scripts/index.js
```diff
@@ -0,0 +1,52 @@
...
+    })
+    }
+  )});
+
+const triggerModals = () => {
```
alicelieutier, line 38: In your implementation, each article has a modal. That means I can open an article on top of another.
Another implementation you could have is that there is only one modal, and it gets filled with whichever is the "current" article.

scripts/news.js
alicelieutier: Totally agree, I really like these.

scripts/spec/newsSpec.js
alicelieutier: This test doesn't really test anything of your code. You set a variable to something, and then test that that variable's value is something. This code above would work even if the News class was completely empty.
The good news is that testing state is not really useful and you can just remove this test.

alicelieutier (commented):
Firstly - love the decision to fall back to full-text when the summary does not work. This shows good product thinking.
It's a common good practice to think of how a page will still provide value in a "degraded state", when one or several things do not work/ will not load, etc.

In terms of UI - I see you used a modal to display the article (not a true modal, actually, as the idea of a modal is normally that background content is disabled, usually by a whole screen overlay div). It's nice to be able to show how you can use modals, but I don't know if it provides the best UI - it's a bit small and the user cannot resize it. Maybe a two column display would work better here? Or a front-end "routing" system based on the hash?

Overall, your code is working well, and I find it easy to read.
I believe you could improve the organisation even more, to separate concerns more. See my comments below.

PR #329, opened 2021-01-24
News Summary Challenge - Lisa Bardelli

lisabardelli:


README.md
```diff
@@ -1,31 +1,23 @@
...
-You'll create an app that summarises the news.
-
-### Guidance
+## Run locally
+In order to run this application locally, please run the following commands:
+```bash
+cd /path/to/the/root/folder/of/this/project
+npm install
+npm start
+```
```
Will-Helliwell, line 13: I followed these instructions but it doesn't seem to work - just a white blank screen with today's date and "News:" subheading

src/interface.js
```diff
@@ -0,0 +1,122 @@
+const guardianAPIKey = 'INSERT_YOUR_API_KEY_HERE';
```
Will-Helliwell, line 1: Good extraction of constants!

src/interface.js
```diff
@@ -0,0 +1,122 @@
...
+function getUrl() {
+  return window.location.hash.split('#')[1];
+}
+
+
+function getNews() {
+  const requestUrl = guardianUrl;
+  hideContent('news_summary');
+  fetch(requestUrl, {
+    method: 'GET',
```
Will-Helliwell, line 39: I'm not sure if it is essential to include lines 39-42 as I think they are the default options. However, it doesn't make it less readable to include them

src/interface.js
```diff
@@ -0,0 +1,122 @@
...
+  const newsUrl = getUrl();
+  if (newsUrl == null || !newsUrl) {
+    getNews();
+  } else {
+    getSummary(newsUrl);
+  }
+};
+
+function getSummary(newsUrl) {
+  hideContent('news_collection');
```
Will-Helliwell, line 26: I really like the overall design - hiding the single article summary and then unhiding it (and hiding the multiple article summary) on a hashchange. All the refactoring you have done makes it very readable and so, even though I can't see it in action, I can see fairly quickly how it works. Well done!

src/interface.js
```diff
@@ -0,0 +1,122 @@
...
+  summarizeText(newsUrl);
+}
+
+function getUrl() {
+  return window.location.hash.split('#')[1];
+}
+
+
+function getNews() {
```
Will-Helliwell, line 35: This function is quite large and I imagine if you had more time you would have refactored out the different elements (seems to me at least: i) fetchNews ii) display news).

src/news.js
```diff
@@ -0,0 +1,13 @@
+class News {
```
Will-Helliwell, line 1: I like that you have refactored some backend logic even though it wasn't part of the instructions - it seems to me that if one ever wanted to expand this application then that would be an essential first step.

Will-Helliwell (commented):
This is really good Lisa - in general very well organised and well refactored. Was very clear to follow. Also, I really like the way you switch between user views by hiding content!

PR #332, opened 2021-01-24
Natasha's attempt at News Summary

natashamcintyre:


interface.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function setTodaysDate() {
+  let todaysDate = new Date().toDateString();
+  document.getElementById('date').innerHTML = `${todaysDate}`;
+}
+
+// This is the function I used to avoid using (and therefore maxing out) the offical API key.
+function getTempHeadlines() {
+  const HEADLINE_REQUEST = `https://content.guardianapis.com/search?from-date=${today}&to-date=${today}&show-fields=thumbnail&api-key=test`;
```
iantfleming, line 46: Good idea to set the API link as a variable

index.html
```diff
@@ -0,0 +1,42 @@
...
+  </head>
+  <body>
+    <!-- Fixed header for title & date -->
+    <div class="header">
+        <h1 id="header-title"></h1>
+        <h3 id="date"></h3>
+    </div>
+    <div class="headerMargin"></div>
+
+    <!-- Pop up window -->
```
iantfleming, line 18: Nice idea to use comments to make it clear what the HTML is doing

index.css
```diff
@@ -0,0 +1,119 @@
...
+}
+
+a {
+  text-decoration: none;
+  color: #0d134a;
+}
+
+.anchorman-gif {
+  text-align: center;
+}
```
iantfleming, line 119: You've clearly put a lot of effort into the styling! I'm sure it looks beautiful.

README.md
```diff
@@ -1,5 +1,58 @@
...
+```
+Worked out how to get text and size to change with respect to window width - use vw in sizing. Eg 10vw means 10% of window width. Works for height too :)
+
+```
+As a busy politician
+I can see whizzy animations in the app
+To make my news reading more fun
+```
+Aminations too complicated for the amount of time I have. Used a gif of Anchorman instead.
+
```
iantfleming, line 55: I like that you have described what you did for each user story.

interface.js
```diff
@@ -0,0 +1,144 @@
...
+  return todaysDate.substring(0, 10);
+}
+
+function getCurrentTime() {
+  let d = new Date();
+  let h = doubleDigits(d.getHours());
+  let m = doubleDigits(d.getMinutes());
+  let currentTime = `${h}:${m}`
+  console.log(`getting time, current time is ${currentTime}`)
+  document.getElementById('header-title').textContent = `Your News Summary at ${currentTime}`
```
iantfleming, line 28: This is a nice feature to have

interface.js
```diff
@@ -0,0 +1,144 @@
...
+    history.replaceState(null, null, ' '); // this removes the hash from the url
+  }
+  window.onclick = function(event) {
+    let gifDiv = document.getElementsByClassName("anchorman-gif")[0];
+    if (event.target == modal || event.target == gifDiv) {
+      modal.style.display = "none";
+      history.replaceState(null, null, ' ');
+    }
+  }
+}
```
iantfleming, line 144: This is pretty extensive! nice work.

iantfleming (commented):
I would say that your code is fairly easily readable (considering it's javascript) though I'm not really sure if it is easy to change or build upon as I am no expert on the matter. It seems like you've set everything up in a way that makes it easy to communicate between the index and the interface so that is definitely a step in the right direction.

PR #333, opened 2021-01-24
Javi's solution

xavierloos:
:warning: Not Implemented: 

```
As a busy politician
I can read the site comfortably on my phone
Just in case my laptop breaks
```
See [Makers Today News](http://makers-today-news.surge.sh/) for a live preview


index.html
```diff
@@ -0,0 +1,20 @@
...
+
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>News</title>
+
+  <link rel="stylesheet" href="public/css/index.css">
+
+  <script defer src="src/news.js"></script>
+  <script defer src="src/interface.js"></script>
```
siellsiell, line 11: good use of defer!

src/news.js
```diff
@@ -0,0 +1,50 @@
...
+    let month = String(today.getMonth() + 1).padStart(2, '0'); //January is 0!
+    let year = today.getFullYear();
+    today = `${year}-${month}-${day}`
+    return today
+  }
+  connectAPI() {
+    let url = "https://content.guardianapis.com/search?"
+    let tag = "&tag=politics/politics"
+    let date = this.getDate()
+    let api = "&api-key=be4ccfea-6ee5-4d67-aa1e-2d8c1a87574e"
```
siellsiell, line 17: It's dangerous to put API keys directly into code you upload to github (or anywhere else). Why do you think that might be and what are ways that you could avoid doing this?

src/news.js
```diff
@@ -0,0 +1,50 @@
...
+      news_text.innerHTML = `${res[i].fields.bodyText}`.substring(0, 300) + "..."
+      let news_link = document.createElement('a')
+      news_link.setAttribute("href", `${res[i].webUrl}`)
+      news_link.innerHTML = "See more"
+      news_div.appendChild(news_img);
+      inner_div.appendChild(news_title);
+      inner_div.appendChild(news_text);
+      inner_div.appendChild(news_link);
+      news_div.appendChild(inner_div);
+      document.body.appendChild(news_div)
```
siellsiell, line 46: You could make this code a little more readable by moving some of those lines into separate functions with meaningful names.

test/newsTest.js
```diff
@@ -0,0 +1,24 @@
+describe("Note Class", () => {
```
siellsiell, line 1: Still says `Note` here! Watch out for copy paste mistakes :)

test/newsTest.js
```diff
@@ -0,0 +1,24 @@
...
+  })
+  describe("connectAPI()", () => {
+    it("is defined", () => {
+      checkIf(news).has("connectAPI").function();
+    })
+  })
+  describe("renderNews()", () => {
+    it("is defined", () => {
+      checkIf(news).has("renderNews").function();
+    })
```
siellsiell, line 22: I'm not sure these two tests are very strong. You're not actually testing the effect of these function at all. Could you make stronger assertions here?

src/news.js
```diff
@@ -0,0 +1,50 @@
+class News {
```
siellsiell, line 1: It feels like you could decompose this into more classes  -- one class I could imagine would be an `Article` class which would know how  to create HTML to display a single article but leaves the fetching of articles to a different part of your code.

public/css/index.css
```diff
@@ -0,0 +1,90 @@
+body {
+  padding: 0;
+  margin: 0;
+  background-image: linear-gradient(to top, #ebbba7 0%, #cfc7f8 100%);
+}
+
```
AmanTank187, line 6: Great styling for your webpage!

src/news.js
```diff
@@ -0,0 +1,50 @@
...
+    fetch(`${url}${tag}&from-date=${date}${api}&show-fields=all`)
+      .then(res => {
+        return res.json()
+      })
+      .then(data => {
+        let res = data.response.results
+        this.renderNews(res)
+      })
+  }
+  renderNews(res) {
```
siellsiell, line 27: You might consider separating the processing of the API results from the creation of DOM elements -- it's hard to write a meaningful test for what this function does otherwise. The recording of Alice's live coding of the Notes app might give some hints for how to do this.

test/newsTest.js
```diff
@@ -0,0 +1,24 @@
+describe("Note Class", () => {
+  var news = new News();
+  it("is defined", () => {
+    checkIf(news).isClass();
+  })
+  describe("getDate()", () => {
+    it("is defined", () => {
+      checkIf(news).has("getDate").function();
+    })
```
siellsiell, line 9: this test is redundant given that the test below would also fail if the function didn't exist.

siellsiell (commented):
Really nice looking app, well done!

- Mobile support: I tried resizing the page using the chrome console and it works well on small screen sizes 💯  You mention that you thought you had to use LocalStorage -- what would you have wanted to use it for? 
- You've combined two user stories a little -- I think you were meant to show the article snippets on a separate page after the user clicks according to the mocks. But this works well too.
- You might want to think about your class abstraction a little more -- it might help you with writing more meaningful tests.

Let me know if you have any questions about any of my comments, happy to go through this in more detail if necessary.

PR #340, opened 2021-02-19
OB new app

bateso88:
Could do with a bit more styling


spec/testingFramework.js
```diff
@@ -0,0 +1,80 @@
+"use strict";
```
louiswicks, line 1: A very extensive and thorough testing framework set up, it looks great, what are the reasons for using listCounters and subListCounters?

interface.js
```diff
@@ -0,0 +1,94 @@
...
+  for(let i =0; i < 10; i++) {
+    addHTMLdiv(`${i}`,'container')
+  }
+}
+
+function printTitles(model) {
+  for(let i =0; i < 10; i++) {
+    addImage(model, i);
+    addHTMLtextElement('h1', `heading-${i}`,`${i}`, model.data[i].webTitle)
+  }
```
louiswicks, line 67: A very clever way of creating multiple dividers and titles for your page and great refactoring them into their own funcitons. That is something I hadn't thought to do!

interface.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+function addImage(model, i) {
+  let parent = document.getElementById(i);
```
louiswicks, line 71: What element are you getting here, and does it not have to be a string?

interface.js
```diff
@@ -0,0 +1,94 @@
...
+  let location = `images/${model.data[i].pillarName}.png`;
+  let image = new Image();
```
louiswicks, line 73: Where are you setting/creating the Image class, or is this an inbuilt class perhaps?

spec/testsSpec.js
```diff
@@ -0,0 +1,12 @@
...
+
+// e.g.
+describeClass('Arithmetic1', function() {
+  describe('Sum', function(){
+    it('3 + 4 equals 6', function() { 
+      expect(3+4).toEqual(6);
+    });
+  });
+});
+
```
louiswicks, line 12: Perhaps could have written more tests for functions? However, not sure if that's needed.

louiswicks (commented):
Great overall code, very methodical commits, and very useful to see clearly split up and structured files. I'll definitely be referring to it when I give the challenge a proper attempt later in the week! Not much to pick apart, maybe more tests?

spec/testingFramework.js
bateso88: these counters just mean that when you create a new HTML list they have a different ID to the last one made. Not essential but thought it could be useful if I ever want to expand the framework

interface.js
bateso88: It's a div with id 'i' (created in line 59). This function itself is called on 65

interface.js
bateso88: in built 👍

PR #343, opened 2021-02-21
News app start Anna

acavalla:


MegGosnell (commented):
Overall i think you did a really great job! Regarding the if(true === false) what happens to your code if you was to take it out?

PR #344, opened 2021-02-22
Partial news summary challenge

OscarB89:
Displays clickable news titles with links to website


HamishArro (commented):
all solid code, I'm not sure what the best practice is for API's but I think it would be better to store them in a separate API file instead of being inside the interface.js file.

PR #346, opened 2021-02-22
News Summary Challenge

Mornevanzyl:
Partial solution. I had problems accessing the Aylien API data from within the Guardian data.


PetarSimonovic:
Code looks very clean and lean Morné. It's impressive how you manage to get so much done with so little. There's a lot here that I'm unfamiliar with so looking forward to going over it with you. Small point but think you need to update the README (still has references to bowling).

PR #351, opened 2021-02-22
News Summary Challenge - Reeshul

reeshul:
Could do with some tests...


index.html
```diff
@@ -0,0 +1,88 @@
...
+          <p class="headline-title" id="m8"></p>
+        </div>
+      </div>
+      <div class="headline-wrapper">
+        <img class="headline-thumbnail" />
+        <div class="info">
+          <p class="headline-title" id="m9"></p>
+        </div>
+      </div>
+    </main>
```
alicelieutier, line 77: This repeating html should probably get created from your JS.

js/interface.js
```diff
@@ -0,0 +1,90 @@
...
+let openModal0 = document.getElementById("m0");
+let openModal1 = document.getElementById("m1");
+let openModal2 = document.getElementById("m2");
+let openModal3 = document.getElementById("m3");
+let openModal4 = document.getElementById("m4");
+let openModal5 = document.getElementById("m5");
+let openModal6 = document.getElementById("m6");
+let openModal7 = document.getElementById("m7");
+let openModal8 = document.getElementById("m8");
+let openModal9 = document.getElementById("m9");
```
alicelieutier, line 13: Is there a way to DRY this code?
Actually, this applies to the whole file.

You could have a loop that:
* creates the element
* sets a click handler to the element
* adds it on the page.

js/script.js
```diff
@@ -0,0 +1,95 @@
...
+  "aylien?apiRequestUrl=https://api.aylien.com/api/v1/summarize?url=";
+const queryImageThumbnail = "?show-fields=thumbnail";
+
+for (let articleId = 0; articleId < 10; articleId++) {
+  fetchNewsArticle(articleId).then(
+    (data) =>
+      (document.getElementsByClassName("headline-title")[articleId].innerText =
+        data.webTitle)
+  );
+  fetchArticleImage(articleId);
```
alicelieutier, line 15: Here you are fetching 10 times for 10 articles. Is there a way to batch this request?

alicelieutier (commented):
Your code is pretty clean overall - it's easy to know what is happening.
I would DRY it in some places.
Also, it's useful to try and limit the number of calls you are making to APIs. Maybe there are more fields you can send the Guardian to get the thumbnail back on the first request?

On testing - with your own testing framework, it can be hard to stub fetch, but not impossible. You know fetch should return an object that has a then property. That then property is a function, that itself returns an object with a then property. And that last then property is a function that takes a callback, and calls is with data.
So potentially, you could have something like:
```js
// before
const saveFetch = fetch;

// where you need to stub
let myMockData = // some data from your examples.
fetch = function() {
  return {
    then: function(){
      return {
         then: function(callback) {
            callback(myMockData)
          }
      }
   }
}
```

js/script.js
Reeshul: I'm not sure I understand what you mean here. Do I not need to fetch 10 article headlines and 10 thumbnails for the page?

Reeshul:
I have cleaned up the index.html and kept the interface logic DRY. Also attempted to stub the api call following your suggestion.

js/script.js
alicelieutier: You do, but you can get the 10 headlines to come back with the url to the thumbnail attached (see thumbnail field on this page: https://open-platform.theguardian.com/documentation/search). It still means you'll have to have more queries for the summary, but that can be delayed to only happen when the user clicks.

js/interface.js
```diff
@@ -1,88 +1,20 @@
...
-  fetchNewsSummary(9);
-};
-
-// When the user clicks on <span> (x), close the modal
-span.onclick = () => {
+// Create 10 headline cards
+// If the user clicks the headline, open the modal and fetch the summary
+for (let i = 0; i < 10; i++) {
+  createHeadlineCards();
+  document.getElementsByClassName("headline-title")[i].onclick = () => {
```
alicelieutier, line 5: This is already much better, but you could do this even more together. When you create directly, you have access to the elements from JS, so no need to search them by class.

alicelieutier: For example you could have:
```js
let openModal = (i) => {
    modal.style.display = "block";
    headlineIntoModalHeader(i);
    fetchNewsSummary(i);
 }
 createHeadlineCards(i, openModal)
```
And inside `createHeadlineCards`:
```js
const createHeadlineCards = (index, clickHandler) => {
  // ...
  const headlineTitle = document.createElement("p");
  headlineTitle.onclick = () => openModal(i)
  // ...
}
```

testing/tests.js
```diff
@@ -1,7 +1,41 @@
...
-    "Headlines page"
-  );
+it("fetches headline thumbnail as expected", () => {
+  mockFetch(myMockData)
+    .then((response) => response.json())
+    .then((data) =>
+      expect(data.response.content.fields.thumbnail).toEqual(
+        "https://media.guim.co.uk/e5517000b6c4c991eff6fff9e496b1af7e52595a/0_330_2448_1469/500.jpg"
+      )
+    );
```
alicelieutier, line 8: this is just testing your mock, not actually testing you code. In order to do that, you have to replace the real fetch with your mockFetch, and then call your functions.

PR #360, opened 2021-03-22
AJ's News Summary Challenge

AJ8GH:
Highlights:
- All user stories complete, except for animations
- Testing framework working well and looking good
- API requests and templating working well, rendering links, headlines, articles and images correctly
- Mobile friendly

To improve:
- Testing front end templating and api methods - improve use of stubbing and quality of these tests
- extend the domain model and use the article object to wrap the article data, instead of working with the JSON response data


README.md
```diff
@@ -1,41 +1,123 @@
...
+- Designed for Politicians using these [user stories](#user-stories)
 
-## Challenge
+## Getting Started
 
-As usual please start by forking this repo.
+To use the app, checkout the deployed version here:
 
-You'll create an app that summarises the news.
+[The News](http://rebel-ocean.surge.sh/)
```
jshorns, line 13: I don't know if these are problems with the hosting, but there were a few glitches for me when running this in the browser. Some of the links to go to the article summaries don't work, and some of them do. I also had one headline showing twice at one point but that may well have been a glitch in the API

README.md
```diff
@@ -1,41 +1,123 @@
...
+
+### interface.js
+
+- Templates the article data into index.html from the JSON responses
+
+```js
+function showHeadlinesAndPictures() {
+  setTimeout(() => {
+    let html = ''
+    for (let i = 0; i < frontPage.articles.length; i ++) {
```
jshorns, line 98: what's the reason for having a counter here rather than forEach or for x of y?

app.js
```diff
@@ -0,0 +1,6 @@
+const httpServer = require('http-server');
```
jshorns, line 1: honestly I did not get the point of http-server this week. I still don't! will have to ask you about this 😂

public/css/style.css
```diff
@@ -0,0 +1,70 @@
...
+  margin-top: 40px;
+}
+
+
+.headline {
+  font-size: 20px;
+  font-weight: 900;
+}
+
+@media (min-width: 1000px) {
```
jshorns, line 40: assuming this is to do with phone display? would love to know how tricky this was. I loaded on my phone and I couldn't see any of the headlines but I was also getting a warning saying the connection wasn't secure, so I think my phone might have been blocking the api requests. Anyway, it fit the display perfectly!

public/index.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
```
jshorns, line 1: it won't let me comment on your gifs above but they're great!

public/js/buffalo/SpecRunner.html
```diff
@@ -0,0 +1,47 @@
...
+          <p class="tagline">Behaviour Driven Development for JavaScript</p>
+          <img class="logo" src="images/favicon-32x32.png">
+        </header>
+        <section id="result-count"></section>
+        <div class="dots" id="dots"></div>
+        <main id="specs"></main>
+      </div>
+
+      <footer id="footer">
+        <p class="footer-text">Testing first, coding later</p>
```
jshorns, line 43: lol

public/js/spec/articleSpec.js
```diff
@@ -0,0 +1,24 @@
+let article;
+
+describe('Article', () => {
+  describe('headline', () => {
+    it('shows the article headline', () => {
+      article = new Article({headline: "Test"});
```
jshorns, line 6: so i had read the project challenge, in saying we weren't to use any backend javascript, to mean that we shouldn't write any classes or anything and that everything should be done from the interface? i might have got the wrong end of the stick there tho.

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
+const POLITICS_URL = `https://content.guardianapis.com/search?from-date=${yesterday()}&to-date=${today()}&q=politics&api-key=test&show-fields=all`
+
+const frontPage = new FrontPage();
+
+fetchArticles();
+
+function fetchArticles(url = POLITICS_URL) {
+  if (ENVIRONMENT === 'test') {
+    const results = '{"response":{"status":"ok","results":"Test OK"}}'
```
jshorns, line 9: nice

public/js/src/frontPage.js
```diff
@@ -0,0 +1,10 @@
+class FrontPage {
+  constructor() {
+    this.articles = []
+    this.summaries = []
+  }
+
+  webUrls() {
+    return this.articles.map(article => article.webUrl)
```
jshorns, line 8: this is v slick

public/js/src/interface.js
```diff
@@ -0,0 +1,45 @@
...
+  .innerHTML = `<img src="${thumbnail}"><br><a href="${webUrl}" target="_blank">${headline}<a>`
+  return frontPage.articles[index].webUrl;
+}
+
+function showSummary() {
+  fetchSummaries(getArticleUrl())
+  setTimeout(() => {
+    document.getElementById('headlines')
+    .innerHTML = `<article class="article">${frontPage.summaries[frontPage.summaries.length - 1].text}</article>`;
+    window.scrollTo(0, 0);
```
jshorns, line 39: ooh what does this do

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
...
+    mode: 'cors',
+    headers: {
+      'Content-Type': 'application/json'
+    }
+  }).then(response => response.json())
+    .then(data => frontPage.articles = data.response.results);
+}
+
+function today() {
+  const today = new Date().toLocaleDateString();
```
jshorns, line 24: I never got beyond using the guardian test api - assuming the api needs you to put in a date range. this is very smart

README.md
AJ8GH: yep definitely is glitchy, I would really like to find out why! Ran out of steam on Sunday to track down the reason

README.md
AJ8GH: Arbitrary decision to go with a for loop, no reason in particular 🙂

app.js
AJ8GH: You and me both 😂

public/css/style.css
AJ8GH: yah, something I found online. Interesting 🤔 I think I needed to deploy it to an http address, maybe that would have fixed the security thing. Thanks for the feedback 🙌

README.md
```diff
@@ -1,41 +1,123 @@
...
+### Testing framework
+
+- In line with the guidelines for the week, there were no frameworks or libraries used in this challenge
+- Testing was done using the framework Buffalo which I wrote in js.
+- The framework reports the number of passes, failures and errors, and shows which tests are passing, giving information about why the failure or error occured
+
+![buffalo-demo](pubilc/images/../../public/images/buffalo-demo.gif)
+
+![buffalo-errors](pubilc/images/../../public/images/errors.gif)
+
```
siellsiell, line 119: Really comprehensive README, nice!

public/js/spec/apiSpec.js
```diff
@@ -0,0 +1,27 @@
+describe('Api', () => {
+  describe('fetchArticles', () => {
+    it('makes fetch request to given url', () => {
```
siellsiell, line 3: not sure that the description is accurate since you're don't have an expectation around what URL the request is being made to

public/js/buffalo/buffalo.js
```diff
@@ -0,0 +1,71 @@
...
+  try {
+    return it();
+  } catch(err) {
+    addErrorToGroup(err.message);
+  }
+}
+
+updateResultCount();
+
+function it(test, expectation) {
```
siellsiell, line 14: the argument names surprised me a little -- I think someting like `function it(test_description, test)` sounds more natural to me as `expectation` makes me think of only the line that says `expect(...)`

public/js/spec/articleSpec.js
siellsiell: Hey Joey! No, in this case it means that you shouldn't write any code that runs on a server rather than in the browser. It's still a good idea to write your code with good separation of concerns and separate your interface from your models even when writing code for the browser. You can think of backend and frontend as synonyms for server and client.

public/js/spec/interfaceSpec.js
```diff
@@ -0,0 +1,8 @@
+describe('Interface', () => {
+  describe('showHeadlinesAndPictures', () => {
+    it('renders the headlines and pictures', () => {
+      // frontPage.articles = [{headline: 'Test', thumb}]
+      // expect(headlines).toInclude('Test')
```
siellsiell, line 5: what happened here? :)

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
+const POLITICS_URL = `https://content.guardianapis.com/search?from-date=${yesterday()}&to-date=${today()}&q=politics&api-key=test&show-fields=all`
+
+const frontPage = new FrontPage();
+
+fetchArticles();
+
+function fetchArticles(url = POLITICS_URL) {
+  if (ENVIRONMENT === 'test') {
```
siellsiell, line 8: this was a good idea for getting around the fact that you couldn't use any libraries for mocking! Generally we want to avoid putting such test-specific behaviour into our code though. One other technique that might have worked is a technique "monkey patching". It would allow you to replace the `fetchArticles` function with a different function in your tests.

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
...
+function yesterday() {
+  const yesterday = new Date();
+  yesterday.setDate(yesterday.getDate() - 1)
+  return yesterday.toLocaleDateString().split('/').reverse().join('-')
+}
+
+function fetchSummaries(articleUrl) {
+  const makersUrl = 'http://news-summary-api.herokuapp.com/aylien?apiRequestUrl='
+  const aylienUrl = 'https://api.aylien.com/api/v1/summarize?url='
+  const url = makersUrl + aylienUrl + articleUrl
```
siellsiell, line 37: since these never change, you could pull them out of the function and have them be top-level constants.

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
...
+
+  if (ENVIRONMENT === 'test') {
+    const result = '{"response":{"status":"ok","results":"Test OK"}}'
+    frontPage.summaries.push(result);
```
siellsiell, line 41: For better encapsulation, code that's outside of the `FrontPage` class shouldn't interact directly with frontPage.summaries. This makes it more onerous to change the internals of `FrontPage` later if you chose to stores the summaries in a different data structure.

And actually ... directly referring to the `FrontPage` class from this file at all makes your API layer code and your model layer code quite intertwined. I think it would be more cleanly separated if you had something like a controller (like you sort have inside `interface.js`) that calls `fetchArticles` and `fetchSummaries` and passes the resulting data into a `FrontPage` class.

public/js/src/api.js
```diff
@@ -0,0 +1,53 @@
+const POLITICS_URL = `https://content.guardianapis.com/search?from-date=${yesterday()}&to-date=${today()}&q=politics&api-key=test&show-fields=all`
+
+const frontPage = new FrontPage();
+
+fetchArticles();
```
siellsiell, line 5: I like that you've separated your fetching code from your models! I wonder whether you could separate this more cleanly from your interface though by putting the actual code that initiates the fetch into your interface.

public/js/src/interface.js
```diff
@@ -0,0 +1,45 @@
+showHeadlinesAndPictures();
+resetHashUrl();
+showSummaryOnClick();
+
+function showHeadlinesAndPictures() {
+  setTimeout(() => {
```
siellsiell, line 6: what's the timeout for?

public/js/src/interface.js
```diff
@@ -0,0 +1,45 @@
...
+    let html = ''
+    for (let i = 0; i < frontPage.articles.length; i ++) {
+      const headline = frontPage.articles[i].fields.headline
+      const thumbnail = frontPage.articles[i].fields.thumbnail
+      html += `<img class="image" src="${thumbnail}"><br>`
+      html += `<a class="headline" href="#${i}">${headline}</a><br>`
```
siellsiell, line 12: Same comment here about violating encapsulation by reaching deep into the internals of `FrontPage`. 

You could make this functionality part of the Article and FrontPage classes, e.g. by adding `toHTML()` methods to each. You might also find that testing is a little easier because you can test the HTML output as part of testing those classes.

public/js/buffalo/interface.js
```diff
@@ -0,0 +1,72 @@
+const main = document.getElementById('specs')
```
siellsiell, line 1: Nice! Looks like you had fun polishing your test framework :)

siellsiell (commented):
This is good solution to the challenge :) You've put some thought into how to separate your interface from your API layer and your models and also put in some nice styling! I left some comments around how you could make the separation even cleaner, which may help you where you struggled with testing. Let me know if you have any questions :)

public/js/src/api.js
AJ8GH: Thanks Simo, this was a tricky area to know what to do. I will bear that approach in mind :)

public/js/src/api.js
AJ8GH: cool thanks, I think I see what you mean but may have to ask you further about this one :) 

Do you mean that `api.js` passes it's data to a separate file / class, say `controller.js`, which then adds it to `frontpage.js`?

public/js/src/api.js
AJ8GH: right, so I think I get what you mean :)

Again just to check I understand correctly, I would still have an `api.js` file, but I would invoke the api request functions from the interface file instead?

public/js/src/interface.js
AJ8GH: I found the API was too slow otherwise and I wouldn't get the articles on the page! If there is a better way to achieve this I would love to know! As I think this may have cause me some bugs, although not sure

public/js/src/interface.js
AJ8GH: Thanks, I definitely agree. I really wanted to extend the classes but ran out of time. Hopefully will find the time to go back and do this

public/js/spec/interfaceSpec.js
AJ8GH: Wasn't sure at the time how to write an effective test for this method. I think maybe with a fresh mentality I might be able to come up with a solution now :)

public/js/buffalo/interface.js
AJ8GH: Thank you Simo 🙂  appreciate the helpful feedback, looking forward to applying your suggestions this week

public/js/src/api.js
siellsiell: Sort of, I mean that in `controller.js` (or in your current `interface.js`) you could have something like:

```
let frontpage = new FrontPage();

fetchArticles(data => frontpage.addArticles(data))
 
# ... code to transform the contents of frontpage into HTML ...
```

where `fetchArticles` is a function inside `api.js`

public/js/src/api.js
siellsiell: Yes :)

public/js/src/api.js
AJ8GH: ah ok, got it, thanks!

PR #363, opened 2021-03-22
news news news

jshorns:


AJ8GH (commented):
- This is really good, especially as you had limited time
- Looks like you got the API calls working and were able to get them on the screen - great work
- Your testing framework looks good
- Your code is easy to follow and read
- One noticeable thing to change is that you committed node_modules folder, meaning you have a ton of additional files in your repo. In future I would stick this in a `.gitignore`
- Other than this you have a great app which meets the core requirements

PR #372, opened 2021-04-19
News Feed single page app vanilla JS

StuBehan:


README.md
```diff
@@ -1,151 +1,20 @@
-# News Summary challenge
```
jackmcc08, line 1: Opened App, 
- It worked on the browser with no tweaks
- All tests passing
- Looks great, simple and workable interface. 
- Liked the testing suite - clearly readable
- I like how you stored the API call in local storage, reduced calls to the API, but how would it work if you want new information? This also avoids the need to mock the test if you run the website first.
- Buttons at top don't do anything - assuming this is a future feature.

README.md
```diff
@@ -1,151 +1,20 @@
...
-If you're interested, you can see the code for the News Summary API in this repo: https://github.com/makersacademy/news-summary-api
-
-## Resources
-
-* [Guardian newspaper API homepage](http://open-platform.theguardian.com/documentation/)
-* [Aylien text summary API docs](http://docs.aylien.com/docs/summarize)
-* cURL [man page](https://curl.haxx.se/docs/manpage.html)
-* [Hurl](https://www.hurl.it/), a web interface for sending HTTP requests
+## The testRunner.html showing testing outcomes
+<img align="center" width="800" src="./docs/news-feed-testing.png">
```
jackmcc08, line 20: Simple readme with clear explanations, I like it!

src/newsItem.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  getBody () {
+    return this.body
+  }
+
+  getPicture () {
+    return this.picture
+  }
+}
```
jackmcc08, line 24: What is the reason for using getter functions? Is this best practice?

src/newsList.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  getList () {
+    return this.list
+  }
+
+  getCounter () {
+    return this.counter
+  }
+}
```
jackmcc08, line 19: news item and news list classes and functions, along with the tests all look good, no comments there.

testRunner.html
```diff
@@ -0,0 +1,45 @@
...
+    <div class="column">
+      <div id='testResults'></div>
+    </div>
+    <div class="column">
+      <div id="indexDisplay"></div>
+    </div>
+  </div>
+  
+</body>
+</html>
```
jackmcc08, line 45: I like how you set up your testing suite, and how it displays the HTML along side it. 

Would have been simpler for me to do the same instead of trying to tear it down after each test!

spec/interfaceSpec.js
```diff
@@ -0,0 +1,8 @@
+window.onload = function() {
+  describe('Interface', function () {
+    it('gets information from the api request and displays them individually', function () {
+      expect(document.getElementById("index").contentDocument.getElementsByClassName("expand").length)
+      .toEqual(list.getList().length)
+    })
+  })
+}
```
jackmcc08, line 8: Does this simply check the number of articles? 

Some thoughts on other tests: 
- Could test if the expand function works 
- Could test if the API call is made to the guardian, i.e. test your fetch function (this could be a nice integration of a mock)

src/interface.js
```diff
@@ -0,0 +1,85 @@
...
+function storeDataLocally(data) {
+  jsonData = JSON.stringify(data)
+  localDataStore.setItem('feedData', `[${jsonData}]`);
+}
+
+async function getLocalData() {
+  let data = await JSON.parse(localDataStore.getItem('feedData'))
+  return await data
+}
+
```
jackmcc08, line 29: is getting info from local Store an async function?

src/interface.js
```diff
@@ -0,0 +1,85 @@
...
+async function getData (url) {
+  let response = await fetch(url)
+  if (response.status !== 200) {
+    console.log(`There was a problem. Status code: ${response.status}`)
+    return
+  }
+  return await response.json()
+} 
+
```
jackmcc08, line 38: what does the second await do?

src/interface.js
```diff
@@ -0,0 +1,85 @@
...
+    expand[i].addEventListener("click", function() {
+      this.classList.toggle("active")
+      var content = this.nextElementSibling
+      if (content.style.maxHeight) {
+        content.style.maxHeight = null
+      } else {
+        content.style.maxHeight = content.scrollHeight + "px"
+      } 
+    })
+  }
```
jackmcc08, line 78: Could you have added in the event listener assigner into the for Each function? 

Also could you extract this function here as well.

src/interface.js
```diff
@@ -0,0 +1,85 @@
...
+}
+
+function displayFeed (feedElement) {
+  if (newsFeed !== null) {
+    newsFeed.insertAdjacentHTML('beforeend', feedElement);
+  }
+}
```
jackmcc08, line 85: Overall the Interface class is very neat, with well extracted classes, I had a couple of questions as per above about how you have written some of the functions.

jackmcc08 (commented):
Overall looks great, with passing tests and the html page looks really slick and smooth functioning. 

Apart from comments in the code, my only thoughts on improvements are you have no way to get a new set of news items without manually eliminating the local storage - although I assume this is more to do with not wanting to overwhelm the API. 

It should also be pretty easy to set up a mock based upon what you have done as well, you just need to create some dummy data.

PR #386, opened 2021-07-19
Chuka's News Summary Challenge

coo990:
Managed to retrieve API data onto webpage but need to work on making it a single-page app. Links go to The Guardian website so I need a way to retrieve headline data into webpage. Applied basic CSS as well as I am still getting familiar with it.


interface.js
```diff
@@ -0,0 +1,17 @@
+// interface.js
+document.addEventListener("DOMContentLoaded", () => {
+  fetch('https://content.guardianapis.com/politics?api-key=41355124-6ce9-43d0-99af-4b99580d961a')
```
YinnyF, line 3: You can use an environment variable to hide api key

index.html
```diff
@@ -0,0 +1,21 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
```
YinnyF, line 3: HTML structure is good, text has tags, indentation is great. 
Maybe the script elements should go in the head?

interface.js
YinnyF: To make this easier to read, you can refactor the steps into its own functions such as `getArticleData()`,  `renderArticleLink(article data)`. 
example: https://github.com/makersacademy/course/blob/main/pills/calling_apis_in_javascript.md

style.css
```diff
@@ -0,0 +1,3 @@
+body {
+  background-color: powderblue;
+}
```
YinnyF, line 3: CSS :D 

One of my friend's advised me to look into CSS Grid to make websites look nicely formatted. There's a really good guide here made by Mozilla if you're interested: https://mozilladevelopers.github.io/playground/css-grid.

YinnyF (commented):
Great work, thanks for sharing Chuka. 
Remember to hide your API Key using an env variable.
Let me know if you check out the CSS Grid guide - I'm currently looking into that too 🤓

style.css
coo990: Thanks! I'm still new to CSS so I thought I would at least try to add some style, I'll definitely look at the link to improve.

interface.js
coo990: I agree it does need to be refactored so it makes more sense

PR #399, opened 2021-11-07
News Summary - Serena Bertozzi

serenabertozzi:
I decided to play with Bootstrap a little bit, so every article is shown inside a card, by clicking the button, the full article appears in a modal. Given more time I would've tried to implement tests and I would've implemented the Aylien API.

### DIFF
--- a/createNewsCard.js
+++ b/createNewsCard.js
@@ -2,7 +2,7 @@ let cardContainer;
 
 createNewsCard = (news) => {
   let col = document.createElement("div");
-  col.className = "col col-md-3 my-4";
+  col.className = "col-12 col-md-6 col-lg-3 my-4";
 
   let card = document.createElement("div");
   card.className = "card shadow cursor-pointer";
--- a/createNewsCard.js
+++ b/createNewsCard.js
@@ -1,0 +1,44 @@
+let cardContainer;
+
+createNewsCard = (news) => {
+  let col = document.createElement("div");
+  col.className = "col col-md-3 my-4";
+
+  let card = document.createElement("div");
+  card.className = "card shadow cursor-pointer";
+  card.style = "width: 18rem; height: 400px;";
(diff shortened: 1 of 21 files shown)


index.html
```diff
@@ -0,0 +1,34 @@
...
+      </div>
+      <hr>
+    </div>
+    
+    <div class="container">
+      <div class="row" id="card-container">
+  </div>
+</div>
+
+<script src="createNewsCard.js" type="text/javascript"></script>
```
starf0g, line 28: You can set up your bundle.js to include these other js files and then only link to the bundle and bootstrap js

createNewsCard.js
```diff
@@ -0,0 +1,44 @@
+let cardContainer;
```
starf0g, line 1: This is really nicely structured and laid out

fetchNews.js
```diff
@@ -0,0 +1,8 @@
+let testAPI =
+  "https://content.guardianapis.com/search?page=1&api-key=test&show-fields=body,headline,thumbnail";
+
+const fetchNews = (onDataFetched) => {
+  fetch(testAPI)
+    .then((response) => response.json())
+    .then(onDataFetched);
```
starf0g, line 7: It's good to add a catch statement here in case there is an error

index.js
```diff
@@ -0,0 +1,23 @@
+// fetchNews((newsData) => {
```
starf0g, line 1: If you require your other JS files here then the bundle.js will include them and you can just use the one file in your HTML (plus the bootstrap js)

starf0g (commented):
This was great! The end result is super functional and looks good.

A couple of things:

- Make sure to add node_modules to your .gitignore so they don't get included in the repo
- Use the bundle.js file so you only have to link one script in your HTML

PR #400, opened 2021-11-07
everything in one commit...

bernardleanse:


bernardleanse:
Found this one hard to do from a TDD and SOLID point of view... Was hard to make divide all the functionality happening following asynchronous fetch into sub functions as when you pass a method of an object into another function, you loose scope of 'this' I eventually learned. I found a way around it after lots of time.

newsFetcher.js
```diff
@@ -0,0 +1,74 @@
+class NewsFetcher {
+  
+
+  getNews(callback){
```
dpkerins, line 4: nice use of callbacks

newsFetcher.js
```diff
@@ -0,0 +1,74 @@
...
+    fetch('https://content.guardianapis.com/search?page=1&api-key=test&show-fields=body,headline,thumbnail')
+    .then(res => res.json())
+    .then(callback);
+  };
+  
+  displayNewsOnPage(data){
+ 
+    data.response.results.forEach(result => {
+    
+      const article = document.createElement('div');
```
dpkerins, line 15: I like the way you split up creating these elements into different functions and then called them together -- your code looks really clean and adjustable.

newsFetcher.js
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  createButton(){
+    const button = document.createElement('button');
+    button.style = "padding: 50px;"
+    button.innerText = 'See Summary';
+    return button;
+  }
+
+  loadSummaryPage(picture, result){
```
dpkerins, line 61: nice simple way to change the body to the article summary

PR #401, opened 2021-11-07
Sergei Palmer PR

s-palmer:
All users stories attempted apart from fancy animations.


src/newsFeed.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+  
+  displayAll() {
+    return this.stories;
+  }
+  
+  findStory(id) {
+    return this.stories[0][id];
+  }
+}
```
ConorButler, line 17: Interesting approach, mapping the response to javascript objects.

ConorButler (commented):
App works perfectly. Some clever approaches to the problem and pretty clean code overall. Could have some more tests but who am I to judge

PR #405, opened 2021-11-08
working webpage

marazzo:
Have only implemented one user story

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,0 +1,45 @@
+const userInput = document.querySelector('.input')
+const newsList = document.querySelector('.news-list')
+const searchForm = document.querySelector('.search')
+
+searchForm.addEventListener('submit',retrieveData)
+
+function retrieveData(e){
+  
+  if(userInput.value == ""){
+    alert('you have not inputted anything')
+  }
+
+  // userInput.value = ""
+  newsList.innerHTML = ""
+
+  e.preventDefault()
+
+  let topic = userInput.value
+  const apiKey = "5a1332c9-b3cf-41b7-a1a3-388cad0b3157"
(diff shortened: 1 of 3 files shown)


app.js
```diff
@@ -0,0 +1,45 @@
+const userInput = document.querySelector('.input')
+const newsList = document.querySelector('.news-list')
+const searchForm = document.querySelector('.search')
+
+searchForm.addEventListener('submit',retrieveData)
+
+function retrieveData(e){
+  
+  if(userInput.value == ""){
```
Pinkish-Warrior, line 9: I love that you have used a new feature as the alert message.

app.js
```diff
@@ -0,0 +1,45 @@
...
+  })
+  .then(resp => {
+    return resp.json()
+  })
+  .then(data =>{
+    console.log(data.response.results)
+    data.response.results.forEach(article =>{
+      let li = document.createElement('li')
+      let a = document.createElement('a')
+      a.setAttribute('href', article.webUrl)
```
Pinkish-Warrior, line 38: It would be a plus if you could use a more meaningful name for the variable "a"; other than that it looks great!!!

index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>News summary</title>
+  <link rel="stylesheet" href="styles.css">
+</head>
+<body>
+
```
Pinkish-Warrior, line 9: something that I have learnt that might be helpful, is to use <div containers> for each part of the section you need to be styling individually.

Pinkish-Warrior (commented):
Very good and great quality code. Well done!!!!!

PR #406, opened 2021-11-08
Incomplete -- News Summary Challenge

dpkerins:


index.js
```diff
@@ -0,0 +1,7 @@
+const News = require('./news');
+
+const news = new News;
+let today = new Date();
+
+news.fetchNewsData(today.toISOString().slice(0, 10));
+
```
bernardleanse, line 7: I like what you've done here, it's going to always get news for whatever date you run the server on!

news.js
```diff
@@ -0,0 +1,39 @@
...
+  createArticleDiv(article) {
+    const mainListDiv = document.querySelector('#main-list');
+    let newArticleDiv = document.createElement('div');
+    newArticleDiv.style = 'padding:10px';
+    newArticleDiv.innerText = article.webTitle;
+    mainListDiv.appendChild(newArticleDiv);
+  };
+  
+}
+
```
bernardleanse, line 36: it's good that you've started this project dividing up this async functionality into 3 methods rather than one with a very long callback which builds the article. Carrying on in this way i'm sure you would have built a well structured program!

news.js
```diff
@@ -0,0 +1,39 @@
...
+      fetch(`http://news-summary-api.herokuapp.com/guardian?apiRequestUrl=http://content.guardianapis.com/search?from-date=${date}`)
+      .then(response => response.json())
+      .then(data => {
+        data.response.results.forEach(article => {
+          this.createArticleDiv(article);
+        })
+      })
+    }
+  };
+
```
bernardleanse, line 17: Finds out how many pages to grab then fetches everything! I can't think of any other way to access how many pages to get. perhaps there might be a simpler way without having to do two calls though.

bernardleanse (commented):
definitely a strong attempt given the circumstances !

PR #416, opened 2021-12-06
Pull request for news-summary-challenge

smoni19:


ginnyamazed (commented):
Nice, clean work.  Easy to follow and understand.   Great start to the project.  I would suspect that if the api's were working you would have also successfully integrated the summary link portion.

PR #419, opened 2022-01-16
Done

tc1316:
Works


bundle.js
```diff
@@ -0,0 +1,105 @@
+(() => {
```
Fadaka, line 1: bundle js is automatic and therefore generally done well, can see scope of how project uses javascript and html

headlines.js
```diff
@@ -0,0 +1,116 @@
+class Headlines {
+  constructor() {
+    this.response = null;
+    this.mainContainerEl = document.querySelector("#main-container");
```
Fadaka, line 4: really good use of constructor variables, initialising, this.elements = null was also quite smart as it solves a-lot of problems later on and I would have also put newsObjects into an array

headlines.js
```diff
@@ -0,0 +1,116 @@
...
+      newsEl.className = "news";
+
+      let newImg = document.createElement("img");
+      newImg.className = "thumbnail";
+      if (item.fields != undefined) {
+        newImg.src = item.fields.thumbnail;
+      }
+
+      this.mainContainerEl.append(newsEl);
+
```
Fadaka, line 32: added. a break div to the display function , interesting

headlines.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+      },
+      false
+    );
+  };
+
+  getSummary = (api_tmp, web_tmp) => {
+    return fetch(`${api_tmp}?show-fields=body&api-key=${API_KEY}`)
+      .then((res) => res.json())
+      .then((res) => {
```
Fadaka, line 88: good use of iterating through array to find text for news articles

Fadaka (commented):
good work in general, code looks like a beaut

PR #429, opened 2022-02-07
Victoria Black - news summary challenge

vblack0:


index.html
```diff
@@ -0,0 +1,23 @@
...
+    <head>
+      <link rel="stylesheet" href="style.css" />
+      <link rel="preconnect" href="https://fonts.googleapis.com">
+      <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+      <link href="https://fonts.googleapis.com/css2?family=Special+Elite&display=swap" rel="stylesheet">
+      <title>News Summary</title>
+    </head>
+    <body style="background-image: url('./images/newspapers.jpg');">
+      <div class='header'>Headlines</div>
+      <div style='text-align: center' class='search'>
```
leoht: A minor one here — the convention is to always use double quotes "" for HTML attributes (even though it'll likely work because the browser handles the single quotes syntax too)

newsModel.js
```diff
@@ -0,0 +1,51 @@
+class NewsModel {
+  constructor() {
+    this.headlines = [];
```
leoht: I'd suggest maybe having a single array, e.g `this.news` in the model, which would be an array of JS objects (either plain key-value objects, or instances of a model class) having each the properties `headline`, `link` and `image`, to make the data more cohesive (and avoiding inconsistent data, for example having the headlines in an order mismatched with the order of the links array, etc)

newsView.js
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  displayNews() {
+    const headlines = this.newsModel.getHeadlines();
+    const links = this.newsModel.getLinks();
+    const images = this.newsModel.getImages();
+    const articlesArray = [];
+
+    headlines.forEach(headline => {
+      const newsEl = document.createElement('div');
```
leoht: Maybe most of this could be wrapped in a private method called `createNewsElement`, so it would simplify the body of the loop — note that having all the elements of a single "news" centralised into a single object (as explained in my comment above) would also simplify the code:
```js
headlines.forEach(article => {
  const newsEl = createNewsElement(article);

  articlesArray.push(newsEl);
});
```

newsView.test.js
```diff
@@ -0,0 +1,66 @@
...
+    };
+    const newsView = new NewsView(newsModel);
+  
+    newsView.displayNews();
+    
+    const headline = document.querySelector('div.headline');
+    expect(headline.querySelectorAll('img').length).toBe(1);
+  });
+
+  it("has a search function", () => {
```
leoht, line 76: To test the search function, you can mock the result returned by the `NewsApi` class when searching with a term — by injecting a "mock" into the view (using the second argument):

```js
const mockedStories = [ { webTitle: 'Extra, extra, read all about it! Test coverage exceeding 100%!!', webUrl: 'https://www.testing.com', fields: { thumbnail: 'testing.png' } }];

// mockApi is just an object with a function getNews() that we're
// mocking - it will call the given callback with our mocked stories

const mockApi = {
  getNews: (term, callback) => {
      callback(mockedStories)
  }
}
const newsView = new NewsView(newsModel, mockApi);

// ...

// setting the search input and clicking the button
searchInput.value = 'random search';
searchButton.click()

Comment (leoht) [reply newsView.test.js:76]: You can also use [Jest automatic class mocks](https://jestjs.io/docs/es6-class-mocks#automatic-mock) (or manual ones), which provide a bit more (you can verify that `NewsApi.getNews()` has been called with the correct search term, mock its behaviour, etc)

File: newsView.test.js [hunk 5]
```diff
@@ -0,0 +1,66 @@
...
+    const newsModel = { 
+      getHeadlines: () => ['testing', 'testing 2'],
+      getLinks: () => [],
+      getImages: () => []
+    }
+    const newsView = new NewsView(newsModel);
+  
+    newsView.displayNews();
+  
+    expect(document.querySelectorAll('div.headline').length).toBe(2);
```
leoht, line 31: Maybe you can make these tests more specific, by adding assertions to verify that the headlines are the correct ones (they should be the values returned by our mocked model above) — same comment for the other tests below, for the other attributes

newsView.js
leoht: Then you could refactor this private method further by adding more methods to construct each element, e.g `createArticleImgElement(article)`, `createArticleLinkElement(article)`, etc

leoht (commented):
Great work! I've added a few suggestions and comments on how to improve the code and more specifically how to simplify the `displayNews()` function, and testing the search function in the `NewsView` tests - let me know if you have any questions on this :)

PR #433, opened 2022-02-07
News-Summary Vanessa Florio

floriov:


headlinesView.js
```diff
@@ -0,0 +1,63 @@
...
+       
+      
+      });
+      this.mainContainerEl.replaceChildren(...newArticles);
+    }
+  
+    
+}
+
+
```
Haydnstokes92, line 60: Code is set out really nice, easily readable and for someone that may not have used the application can see exactly what each element is doing, nice!

newsApi.js
```diff
@@ -0,0 +1,32 @@
...
+    searchHeadlines(searchTerm = '', callback) {
+      fetch(`https://content.guardianapis.com/search?q=${searchTerm}&query-fields=headline&show-fields=thumbnail,headline,byline&order-by=newest&api-key=${API_KEY}`)
+        .then(response => response.json())
+        .then(data => {
+          console.log(data.response.results),
+          callback(data.response.results)})
+        .catch((error) => console.log(error))
+   }
+    
+
```
Haydnstokes92, line 27: Curious as to which resources you used when implementing your api's! really nice work.

PR #435, opened 2022-02-07
news-summary-challenge

robinucar:


bundle.js
```diff
@@ -0,0 +1,132 @@
+(() => {
```
alicelieutier, line 1: I don't think you should commit this file to github. Among other, it contains an API key.

index.html
```diff
@@ -0,0 +1,24 @@
...
+    <input type="text" id='search-input' class='search-input' placeholder="">
+    <button id='search-news' class='search-news'>Search News</button>
+    
+    <section id ='news-list' class="news-list">
+        
+
+        <!--Data appears here-->
+
+    </section>
+    <script src="bundle.js"></script>
```
alicelieutier, line 22: I'm curious about why you bundle your files in one rather than adding them all to the page.

index.js
```diff
@@ -0,0 +1,17 @@
+const NewsApi = require('./newsApi')
+const NewsModel = require('./newsModel')
+const NewsView = require('./newsView')
```
alicelieutier, line 3: In JS in the browser, you don't really need to require files, as the html file is the one that makes sure all the files are there. 
Also, in your case you are bundling them into one file, meaning there's even less need to require. But maybe I'm missing something?

newsApi.js
```diff
@@ -0,0 +1,27 @@
...
+      const response = await fetch(url);
+      const data = await response.json();
+
+      callback(data);
+    } catch (error) {
+      console.error(error);
+    }
+  }
+
+  async searchNews(str = "", callback) {
```
alicelieutier, line 15: naming - the argument `str` should have a more meaningful name, maybe `queryString` ?

newsApi.js
```diff
@@ -0,0 +1,27 @@
+const API_KEY = require("./apikey");
+class NewsApi {
+  async loadNews(callback) {
```
alicelieutier, line 3: This function and the one below are exactly the same, except without the string argument.
Using the DRY (don't repeat yourself) principle here, you could delete this one and use the second one leaving the str argument empty.

newsApi.js
alicelieutier: Also - default parameters only really work when they are the last parameter, otherwise it's not possible to leave the argument blank to be replaced by the default.

newsModel.test.js
```diff
@@ -0,0 +1,26 @@
...
+               "thumbline": "test_img 2",
+           },
+           "webUrl": "testurl_2"
+       }
+    ]
+  const model = new NewsModel();
+  it("should keep  news headlines within array", () => {
+   model.setNews(newsList)
+   expect(model.getNews()).toEqual(newsList)
+  });
```
alicelieutier, line 25: Your class doesn't have behaviour, it's only a structure to store things, I wouldn't test such a class.

index.js
```diff
@@ -0,0 +1,17 @@
...
+const api = new NewsApi()
+const model = new NewsModel()
+const view = new NewsView(model, api)
+
+api.loadNews((data) => {
+    model.setNews(data.response.results);
+    view.displayNews()
```
alicelieutier: ->> HERE <<-- (reference for a comment later)

newsView.js
```diff
@@ -0,0 +1,67 @@
...
+      const header = document.createElement("a");
+      header.innerText = news.fields.headline;
+      header.href = news.webUrl;
+      header.className = "headline";
+      const image = document.createElement("img");
+      image.src = news.fields.thumbnail;
+      image.className = "image";
+      div.append(image);
+      div.append(header);
+      this.mainContainerEl.append(div);
```
alicelieutier, line 53: This is the cause of your undefined bug. More specifically, `this.mainContainerEl` is only defined when one of the buttons above is clicked, even though you call this function here when the page loads which I marked with `->> HERE <<--`

newsView.js
```diff
@@ -0,0 +1,67 @@
...
+      div.className = "news-item";
+      const header = document.createElement("a");
+      header.innerText = news.fields.headline;
+      header.href = news.webUrl;
+      header.className = "headline";
+      const image = document.createElement("img");
+      image.src = news.fields.thumbnail;
+      image.className = "image";
+      div.append(image);
+      div.append(header);
```
alicelieutier, line 52: This is perfectly acceptable and many devs will prefer it but if you want another way to write this code , you could also write this:
```js
const div = document.createElement("div");
div.className = "news-item";
div.innerHTML = `
  <img src="${news.fields.thumbnail}" class="image">
  <a href="${news.webUrl}" class="headline">${news.fields.headline}</a>
`
```

newsView.test.js
```diff
@@ -0,0 +1,47 @@
...
+
+ describe("NewsView class", () => {
+     it('displays news title , url and image after click load button', () =>  {
+        document.body.innerHTML = fs.readFileSync("./index.html");
+         const model = new NewsModel()
+         const view = new NewsView(model)
+         const loadButtonEl = document.querySelector('#load-news')
+            loadButtonEl.click()
+            view.displayNews()
+         expect(document.querySelectorAll('#news-list').length).toEqual(1)
```
alicelieutier, line 35: I'm not sure what you are trying to test here.
Are you testing that when you call displayNews the news are displayed?
OR, are you trying to test that when the button is click the news are displayed?

The second test would be an interface test whereas the first test would test the method inside the view... and actually would fail, revealing the bug pointed above ;)

newsView.test.js
```diff
@@ -0,0 +1,47 @@
...
+ const fs = require('fs');
+ const NewsView = require('./newsView')
+ const NewsModel = require('./newsModel')
+ const NewsApi = require('./newsApi')
+
+ const newsList = [
+     {
+        "fields": {
+            "headline": "Test Headline 1",
+            "thumbline": "test_img 1",
```
alicelieutier, line 14: this should probably read "thumbnail"

alicelieutier (commented):
Good effort, nice separation of concerns.
Also, good use of callbacks!
There are a few logic problems, and you are slightly sidestepping the async complexity of the project by using await.

PR #440, opened 2022-04-10
Jonny's News thing

jonnyabrams:
First four user stories catered for, I think


newsApi.js
```diff
@@ -0,0 +1,26 @@
+const guardianApiKey = require('./apiKey');
```
Lorn888, line 1: smart to exclude your Api key from committing to github :-P

Lorn888 (commented):
Impressive code, especially your TDD !

PR #441, opened 2022-04-11
Valentina's News Summary Challenge

valentina-maggio:
Completed the first 4 user stories and added CSS styling.

I had some troubles mocking the second method for the Guardian API, could you please give me some feedback on this? Also, I wasn't sure on how to store the API key (I tried to use Dotenv, but apparently it only works for the backend), so I decided to create a JS class for that, but I'm not sure that's the best way to do it...

Many thanks :)


README.md
```diff
@@ -1,39 +1,41 @@
...
-HTML. You won't use Ruby or backend JavaScript.
+### Note on API credentials
 
-You're strongly encouraged to use the same tools and libraries you've been learning this
-week:
- * Jest to write your tests
- * The [`jest-fetch-mock`](https://www.npmjs.com/package/jest-fetch-mock) module (or
-   similar) to mock `fetch` requests (so you don't request the API when running tests!)
- * A build tool such as `esbuild` to bundle files together
+The credentials are not uploaded in the current folder. Please use the guardianApiKeyTemplate.js to save your API key and follow the instructions saved in the file.
```
alicelieutier, line 27: This is good enough. Given that this projects is front end, the user would have access to the API key anyway, but at least it's not on Github. What you can do to when using an API key on the front-end is tell the API provider which domains it should authorise requests from, so at least people are not copying the API key into their own projects.

guardianApi.js
```diff
@@ -0,0 +1,26 @@
...
+    fetch(this.url)
+      .then(response => response.json())
+      .then(data => {
+        console.log(data.response.results),
+        callback(data.response.results) 
+      })
+      .catch((error) => console.log(error));
+  }
+
+  searchHeadlines(searchField, callback) {
```
alicelieutier, line 20: This is not a great use of the instance variable as you are first chaging it, then calling another method that uses it.
Because of that, there's no way you can get headlines again after you've done a search, for example.

You may want to make the getHeadline method take a url as a parameter, or have two separate methods that fetch, each with its own url.

newsSummaryModel.js
```diff
@@ -0,0 +1,20 @@
...
+    this.headlines = [];
+  }
+
+  getNews() {
+    return this.headlines;
+  }
+
+  addNews(headline) {
+    this.reset();
+    this.headlines = headline;
```
alicelieutier, line 12: You don't really need a reset since you are overwriting the whole headline array here anyway.

newsSummaryView.js
```diff
@@ -0,0 +1,56 @@
...
+      this.guardianApi.searchHeadlines(`${this.inputSearch.value}`, newsData => {
+        this.model.addNews(newsData);
+        this.inputSearch.value = '';
+        this.displayNews();
+      });
+    });
+  }
+
+  displayNews() {
+    document.querySelectorAll('.headline').forEach(el => el.remove());
```
alicelieutier, line 28: You can do this faster with something like `this.mainContainer.innerHtml = ''`

newsSummaryView.test.js
```diff
@@ -0,0 +1,78 @@
...
+const NewsSummaryView = require('./newsSummaryView');
+
+let newsModel;
+let newsView;
+let fakeApi;
+
+beforeEach(() => {
+  document.body.innerHTML = fs.readFileSync('./index.html');
+
+  fakeApi = { getHeadlines: (callback) => callback(
```
alicelieutier, line 16: Are you declaring a global variable on purpose here?

newsSummaryView.test.js
```diff
@@ -0,0 +1,78 @@
...
+        webUrl: 'https://examplewebsite.com',
+        fields: {
+          thumbnail: 'https://image.jpg'
+        } 
+      }]
+    )
+  };
+
+  newsModel = new NewsSummaryModel();
+  newsView = new NewsSummaryView(newsModel, fakeApi);
```
alicelieutier, line 28: Asme - there are global variables. You may want to use `let` or `const`. Also, adding `"use strict"` at the top of your file will stop you from having globals.
Check https://www.w3schools.com/js/js_strict.asp#:~:text=Strict%20mode%20changes%20previously%20accepted,accidentally%20create%20a%20global%20variable.

newsSummaryView.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+  //   const newsView2 = new NewsSummaryView(newsModel, fakeApi2);
+
+  //   const inputEl = document.querySelector("#search-input");
+  //   const buttonEl = document.querySelector("#search-button");
+
+  //   newsView2.displayNews();
+
+  //   inputEl.value = 'today';
+  //   buttonEl.click();
```
alicelieutier, line 73: It's hard to know how to help, because I'm not sure what was not working. However, I can say that typically click will not work in this type of tests. Is that the problem you faced?

alicelieutier (commented):
Well done!

You fetched, you understood callbacks, you stubbed an API, you hid the API key... You should be proud of yourself 👍 

Be careful with you variable scopes. Variables should be as "local" as possible. Only using instance variables when really necessary, and globals... as little as possible.

I know you were unsure about your mocking... unfortunately it's hard to know what was not working, and without that, it's hard to help..

newsSummaryView.test.js
valentina-maggio: So my idea was to use the beforeEach() to avoid repeating the same lines of code in multiple it blocks, but I found out that if I declared the variables within the beforeEach() block they would be only visible in the scope of that block. Therefore I used the solution of declaring them before the beforeEach block (which I saw in some threads online). Was it a bad idea?

newsSummaryView.test.js
valentina-maggio: Same reason here. I declared the variables outside of the beforeEach block so then they could be visible in every it block. I guess next time it will be better to keep them within the it blocks. But what if I have several it blocks using all the same variables?

newsSummaryView.test.js
valentina-maggio: The problem was actually in lines 61/62. I was trying to create the double fakeApi with the two methods getHeadlines and searchHeadlines but it wasn't working. Perhaps if you are in the office tomorrow and have 5 minutes we can have a look at it together :)

README.md
valentina-maggio: Good tip! Thank you :)

PR #454, opened 2022-05-09
News Summary- Challenge

R552-beep:


index.js
```diff
@@ -0,0 +1,12 @@
+const NewsModel = require('./newsModel')
+const NewsView = require('./newsView')
+
+const model = new NewsModel();
+const view = new NewsView();
```
almorcrette, line 5: You need to include `model` that you created on line 4 in as the argument when creating `view` on line 5.

Without it, when I open `index.html` in the browser, it doesn't show anything because line 12 of this file (`index.js`) doesn't run properly because line 8 of `newsView.js` doesn't know what model to read.

almorcrette (commented):
Hi Ruzeka,

This work looks good. It looks like going through the steps from the week in the weekend challenge has helped you understand and apply the week's learning and skills.

Highlights from this submission:

- [x] Fully test driven - full test coverage
- [x] Production code working as desired
- [x] Good separation of concerns in the file structure

There is a small issue in your main file which prevents the HTML loading as expected. You need to include the model you instantiate as an argument when instantiating the view. I have put a detailed comment in the code review about this. Once I did this, the webpage loads with the expected content.

I recommend that you get in the habit of updating the README to make it your own, including providing instructions to a user on how to get started with your code.

Your next steps would be to replace the hardcoded strings with data fetched from the api. This would involve a method that looks like this (replacing `api-key=test` at the end of the URL with your key.

function:
```
  fetchStories(cb) {
    fetch('https://content.guardianapis.com/search?q=&query-fields=headline&show-fields=thumbnail,headline,byline&order-by=newest&api-key=test')
      .then(response => response.json())
      .then(data => {
        console.log("Success: ", data);
        cb(data)
      });
  }
```

You could then create another function that takes the data received from the API call to save it into your model:

```
  saveStories(data) {
    this.stories = []
    data.response.results.forEach((result) => {
      this.stories.push({
        "webTitle": result.webTitle,
        "webUrl": result.webUrl,
        "thumbnail": result.fields.thumbnail
      });
    });
  };
```
Your existing `view.displayNews` method would then display the news items in the model that have been fetched via the api.

Tests for those two functions, you can see on my submission if helpful. Things to think about with the tests:

- The test for the fetchStories() doesn't actually call the API, it just tests that it takes a callback function as an argument, and returns the outcome of the data being passed through the call back. So you need to set up a stub that imitates the data that the api would fetch.
- The test for saveStories also uses the stub (the imitated data form the api) and checks that it saveStories takes that data and adds it to the models array.

I hope this is useful guidance!

PR #455, opened 2022-05-09
New Summary challenge

NBenzineb:


src/newsView.test.js
```diff
@@ -0,0 +1,48 @@
+
```
mcsuGH, line 1: need to add 
```
/**
 * @jest-environment jsdom
 */
```
for it to run jest properly for test

mcsuGH (commented):
Good testing - used mocks for Api, every method has a corresponding test, and all tests pass (once tested inside jsdom test environment).

Functionality seems like it would work from looking at the View class - just need to fix the fetch on Api and it should be fine.

